PR #6, opened 2022-07-18
Forgot/ResetPassword

zahirekrem09:


libs/ui/src/components/ForgotPasswordForm/ForgotPasswordForm.stories.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+const Template: Story<ForgotPasswordFormProps> = (args) => {
+  const [isLoading, setIsLoading] = React.useState<boolean>(false);
+  const onSubmit = (data: FieldValues) => {
+    setIsLoading(true);
+    setTimeout(() => {
+      setIsLoading(false);
+    }, 2000);
+
+    confirm(JSON.stringify(data));
```
7alip, line 20: I get Eslint error. I wonder if you can also see the warnings. Otherwise we might need to activate somehow locally and prevent pushing with lint errors

libs/ui/src/components/ForgotPasswordForm/ForgotPasswordForm.tsx
```diff
@@ -0,0 +1,108 @@
...
+const schema = (t: any) =>
+  yup.object({
+    email: yup
+      .string()
+      .email(t`contact.form.email-invalid`)
+      .required(t`login.email.required`),
+  });
+
+export type ForgotPasswordFormProps = {
+  onSubmitHandler: (data: FieldValues) => void;
```
7alip, line 18: We can declare `ResetPasswordFieldValues` to be used instead of `FieldValues`, so that we will able to see the field value types in the pages/components where we call onSubmit handler.

For example Strapi expect id as number but we might get the id as string from some forms. Type checking will guide us about type conversions in this case.

```jsx
export type ResetPasswordFieldValues = {
  password: string;
  passwordConfirmation: string;
};
```

libs/ui/src/components/ForgotPasswordForm/ForgotPasswordForm.tsx
```diff
@@ -0,0 +1,108 @@
...
+    },
+    resolver: yupResolver(schema(t)),
+    mode: 'all',
+  });
+
+  const onSubmit: SubmitHandler<FieldValues> = (data) => {
+    onSubmitHandler(data);
+  };
+
+  //   const handleSubmitSign = async (data) => {
```
7alip, line 45: We can remove comments

7alip (commented):
Thank you. I left a small comments about form fields type definition and removing commented out lines.

libs/ui/src/components/ForgotPasswordForm/ForgotPasswordForm.tsx
zahirekrem09: this part was like you said normally, but since there were type errors in the `errors`  and  `register` props in the `FormItem`, I had to do it this way.

```jsx
export type ResetPasswordFieldValues = {
  password: string;
  passwordConfirmation: string;
};

 const {
    register,
    handleSubmit,
    formState: { errors },
  } = useForm<ResetPasswordFieldValues>();
```

libs/ui/src/components/ForgotPasswordForm/ForgotPasswordForm.stories.tsx
zahirekrem09: Unfortunately, I don't see any eslint errors.

libs/ui/src/components/ForgotPasswordForm/ForgotPasswordForm.tsx
7alip: @zahirekrem09 Well, it makes sense then.

You can. maybe fix the lint issues and then merge it.

`yarn lint` or `npx nx affected --target=lint --fix`

7alip (approved):

PR #16, opened 2022-07-23
SignupForm

MustafaTRKYLMZ:
signup form added


libs/ui/src/components/SignupForm/SignupForm.tsx
```diff
@@ -0,0 +1,167 @@
...
+      ),
+    email: yup
+      .string()
+      .email(t`contact.form.email-invalid`)
+      .required(t`login.email.required`),
+  })
+
+export const SignupForm: React.FC<SignupFormProps> = ({
+  onhandleSubmitSignUp,
+  handleTermsAccepted,
```
7alip, line 47:
```suggestion
  onSignup,
  onAcceptTerms,
```

We can use shorter descriptive prop names for the components.

7alip (commented):
Thank you! I have just some small comments

PR #20, opened 2022-07-25
 FileUploader 

zahirekrem09:


libs/ui/src/components/FileUploader/FileUploader.tsx
```diff
@@ -0,0 +1,142 @@
...
+          id="form-input"
+          accept="image/png, image/jpeg, image/jpeg, image/webp"
+          type={'file'}
+          multiple
+          display={'none'}
+          ref={inputRef}
+          onChange={onInputChange}
+        />
+        <VStack
+          h={250}
```
7alip, line 90: Could you try to make the uploader to be able to be stretched to the parent component. Maybe minH={250} and h='full'.

Maybe you can add a story for that as well.

Screenshot 2022-07-25 at 16 33 23
.

7alip (commented):
Great, Thank you!

7alip:
@zahirekrem09, I tested it with minH changes, it doesn't seem to work. Reverted the changes. You can merge it.

PR #24, opened 2022-07-27
CommentItem

yhtyyar94:
there was no createdAt in comment.ts so added
to the user.ts added avatar property


7alip (changes requested):
Looks ok.

Keep in mind, we have updated the types in dev branch. Merging `dev` into `comment-item` should also fix the `createdAt` issue.

PR #25, opened 2022-07-27
CommentList

yhtyyar94:
commentItem needs to be exported from index.tsx


libs/ui/src/components/shared/commentList.tsx
```diff
@@ -0,0 +1,27 @@
+import { Stack, Text } from '@chakra-ui/react';
+import { Comment } from '@wsvvrijheid/types';
+import { useTranslation } from 'next-i18next';
+import { FC } from 'react';
+import { CommentItem } from '~components';
```
7alip, line 5: Let's wait until `CommentItem` in #24 be merged.

7alip (changes requested):
We can review it again after #24 

Component and folder names should be PascalCase, `CommentList` and we don't use subfolder in components like `shared`.

1. Please move the component in `src/components/CommentList/CommentList.tsx`
2. Please create a story for it `CommentList.stories.tsx`

PR #26, opened 2022-07-27
Card

yhtyyar94:
pls check how i imported Navigate comp.
import { Navigate } from '~components'  this method didnt work


libs/ui/src/components/shared/card.tsx
```diff
@@ -0,0 +1,77 @@
+import { Avatar, Center, Heading, Image, Stack, Text } from '@chakra-ui/react';
+import { FC } from 'react';
+import { Navigate } from '../Navigate';
+
+interface CardProps {
+  title: string;
+  description: string;
+  image: string;
+  link: string;
+  rounded: boolean;
```
7alip, line 10:
```suggestion
  rounded?: boolean;
```

It could be optional

libs/ui/src/components/shared/card.tsx
```diff
@@ -0,0 +1,77 @@
...
+              src={process.env['NX_PUBLIC_URL'] + image}
+              transition="transform 0.5s ease-in-out"
+              _groupHover={{ transform: 'scale(1.1)' }}
+            />
+          ) : (
+            <Image
+              objectFit="cover"
+              h={48}
+              w="full"
+              src={process.env['NX_PUBLIC_URL'] + image}
```
7alip, line 45:
```suggestion
              src={image}
```

We could provide absolute image url when we use the component. Otherwise we will only be able to use local images.

7alip (changes requested):
Good job, thanks. 

1. Can you put all the components in `components` folder without creating a subfolder like `shared`? `libs/ui/components/Card/{Card.tsx, Card.stories.tsx, index.tsx}.
2. Can you please create a `Card.stories.tsx` so that we can easily test the component visually.

PR #29, opened 2022-07-30
PlatformList

7alip:
We can create stories for app specific components after we configure storybook in the apps (foundation).


MustafaTRKYLMZ (approved):
If we don't need stories, everything looks good.

PR #31, opened 2022-07-30
CommentForm

MustafaTRKYLMZ:
comment form added


libs/ui/src/components/CommentForm/CommentForm.tsx
```diff
@@ -0,0 +1,133 @@
...
+import { FiArrowRight } from 'react-icons/fi'
+import * as yup from 'yup'
+
+import { FormItem } from '../FormItem'
+import { CommentFormProps } from './types'
+
+import { mutation } from '~lib'
+import { toastMessage } from '~utils'
+
+const userSchema = (t: TFunction<'translation', undefined>) =>
```
7alip, line 24:
```suggestion
const userSchema = (t: TFunction) =>
```

You probably forgot to import TFunction

`import { TFunction } from 'next-i18next'`

We only have `common` as translation files. It could be `TFunction<`common`>` if you import from `react-i18next` or only `TFunction`

libs/ui/src/components/CommentForm/CommentForm.tsx
```diff
@@ -0,0 +1,133 @@
...
+              errors={errors}
+              {...useBreakpointValue({ base: { rows: 1 }, sm: { rows: 3 } })}
+            />
+            <IconButton
+              display={{ base: 'flex', sm: 'none' }}
+              colorScheme="blue"
+              aria-label="Send Comment"
+              icon={<FiArrowRight />}
+              isRound
+              isLoading={isLoading}
```
7alip, line 113: You need to provide `isLoading` from props.

libs/ui/src/components/CommentForm/CommentForm.tsx
```diff
@@ -0,0 +1,133 @@
...
+  auth,
+  onSendForm,
+}) => {
+  const { t } = useTranslation()
+
+  const {
+    register,
+    handleSubmit,
+    formState: { errors, isValid },
+  } = useForm({
```
7alip, line 49:
```suggestion
  } = useForm<CommentFormFieldValues>({
```

It's better to give field values types to useForm

libs/ui/src/components/CommentForm/CommentForm.stories.tsx
```diff
@@ -0,0 +1,26 @@
+import { Story, Meta } from '@storybook/react'
+
+import { CommentForm } from '.'
+import { CommentFormProps } from './types'
+import { CommentFormFieldValues } from './types'
```
7alip, line 5:
```suggestion
import { CommentFormFieldValues, CommentFormProps } from './types'
```

libs/ui/src/components/CommentForm/types.ts
```diff
@@ -0,0 +1,16 @@
+import { User } from '@wsvvrijheid/types'
+export type CommentFormFieldValues = {
+  email: string
+  name: string
```
7alip, line 4:
```suggestion
  email?: string
  name?: string
```

email and name should be optional as we don't need for authenticated users.

7alip (changes requested):
Thank you. I made som comments which are required for build.

libs/ui/src/components/CommentForm/types.ts
MustafaTRKYLMZ: Do we're accepting comments without name or email ?

libs/ui/src/components/CommentForm/CommentForm.stories.tsx
7alip, line 5: It doesn't seem to be resolved!

7alip (approved):
Thank you

PR #43, opened 2022-08-08
Storybook React Query Wrapper

7alip:
This will allow us to use queries and mutations in stories. An example included as `Activity.stories.tsx`. You need to select `Turkish` language from storybook toolbar when testing example Activity component.

Screenshot 2022-08-08 at 19 11 47


MustafaTRKYLMZ (approved):
Thanks 
I think Ard Cardbase diffirent issue. Is that correct?

7alip:
I just renamed ArtCard, didn't change anything. ArtCardBase is just for visual testing and takes the functions via props. But ArtCard will contain data logic. As I mentioned before, it's better to use this approach if the component is supposed to be rendered in a list.

PR #45, opened 2022-08-09
Feat/admin news card

zahirekrem09:
https://github.com/orgs/wsvvrijheid/projects/3/views/1?layout=board


libs/ui/src/admin/NewsCard/NewsCardBase.tsx
```diff
@@ -0,0 +1,98 @@
...
+        height={isVertical ? '150px' : '100%'}
+        objectFit="cover"
+        objectPosition="center"
+        marginLeft={isVertical ? '0px' : 'auto'}
+        borderTopRightRadius={'md'}
+        borderTopLeftRadius={isVertical ? 'md' : '0'}
+        borderBottomRightRadius={isVertical ? '0' : 'md'}
+        src={news.image}
+        alt={news.title}
+      />
```
7alip, line 95: If the card container has a fixed height in horizontal mode (like you set 208/150), the content can overflow in case there are multiple lines of header or description. 

I didn't know it's challenging to make the image fills the auto height flex container in horizontal mode. Finally I found a solution like this.

```tsx
<Image
  w={isVertical ? "full" : "300px"}
  h={isVertical ? "150px" : "auto"}
  objectFit="cover"
  src={news.image}
  alt={news.title}
/>;
```

You don't need border radius for image, just give `overflow="hidden"` to card.

libs/ui/src/admin/NewsCard/NewsCardBase.tsx
```diff
@@ -0,0 +1,98 @@
...
+            <Text>-</Text>
+            <Text> {news.owner} </Text>
+          </HStack>
+          <HStack spacing={4} fontSize="12px" color={'blue.500'}>
+            <ActionButton
+              onClick={() => alert(news.url)}
+              title="View"
+              isVertical={isVertical}
+              icon={AiOutlineEye}
+            />
```
7alip, line 61: I think it's better to use ButtonGroup component for stacked buttons. With removing `iconSpacing` you can easily convert Button to IconButton appearance.

```tsx
<ButtonGroup
  border={0}
  variant="ghost"
  colorScheme="primary"
  size={isVertical ? "md" : "sm"}
>
  <Button
    aria-label='View'
    onClick={() => alert(news.url)}
    leftIcon={<AiOutlineEye />}
    {...(isVertical ? { iconSpacing: 0 } : { children: "View" })}
  />
...
```

libs/ui/src/admin/NewsCard/NewsCardBase.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+export const NewsCardBase: FC<NewsCardProps> = ({
+  isFeatured,
+  isVertical,
+  news,
+}) => {
+  return (
+    <Stack
+      w={isVertical ? '300px' : '1000px'}
+      h={isVertical ? 'auto' : isFeatured ? '208px' : '150px'}
```
7alip, line 18: I would remove width and height. We want the card fill the container in smaller devices.

libs/ui/src/admin/NewsCard/NewsCardBase.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { HStack, Image, Stack, Text, VStack } from '@chakra-ui/react'
+import { AiOutlineEye, AiOutlineLike, AiOutlineShareAlt } from 'react-icons/ai'
+import { BsBookmarkHeart } from 'react-icons/bs'
+
+import ActionButton from './ActionButton'
+import { NewsCardProps } from './types'
+
+export const NewsCardBase: FC<NewsCardProps> = ({
+  isFeatured,
+  isVertical,
```
7alip, line 12: I would give variant `(horizontal | vertical)` instead of boolean value.

```suggestion
  variant = "vertical"
```

In the issue there is a responsiveness requirement:

- [ ] Horizontal variant should always look like vertical variant in small viewports.

```tsx
const isVertical = useBreakpointValue({
  base: true, // Always vertical in smaller viewports
  lg: variant === 'vertical' ? true : false,
})
```

Or the other way around

```tsx
const isHorizontal = useBreakpointValue({
  lg: variant === 'horizontal' ? true : false
})
```

libs/ui/src/admin/NewsCard/types.ts
```diff
@@ -0,0 +1,15 @@
+import { News } from '@wsvvrijheid/types'
+import { IconType } from 'react-icons'
+
+export type NewsCardProps = {
+  isFeatured?: boolean
+  isVertical?: boolean
```
7alip, line 6:
```suggestion
  variant?: "horizontal" | "vertical"
```

7alip (changes requested):
Thanks for great effort.

Required changes are especially for responsiveness and auto height horizontal card. But other notes are optional.

libs/ui/src/admin/NewsCard/NewsCardBase.tsx
```diff
@@ -0,0 +1,114 @@
...
+  news,
+  variant = 'horizontal',
+}) => {
+  const isVertical = useBreakpointValue({
+    base: true, // Always vertical in smaller viewports
+    lg: variant === 'vertical' ? true : false,
+  })
+  return (
+    <Stack
+      w={isVertical ? '300px' : '1000px'}
```
7alip, line 29:
```suggestion
```

I wouldn't limit the width of the card. We don't use vertical card as a single column in large screens. So it's better to have full width component. You can remove w here

libs/ui/src/admin/NewsCard/NewsCardBase.tsx
```diff
@@ -0,0 +1,114 @@
...
+      h={isVertical ? 'auto' : isFeatured ? '208px' : '150px'}
```
7alip, line 30:
```suggestion
      h={isVertical ? 'auto' : '200px'}
```
Card content will overflow if we have multiple lines of header and description with this height. Maybe we can limit description text lines on horizontal mode:

`<Text noOfLines={isVertical ? 3 : 2}>{description}</Text>`

7alip (changes requested):

7alip:
Hey @zahirekrem09! If you don't mind I touched your code, updated some design and added Grid to stories.

Could you merge it please if you agree with the changes?

PR #47, opened 2022-08-11
Feature/new component

MSKose:
I've added the demo for StackExample


libs/ui/src/admin/StackExample/StackExample.stories.tsx
```diff
@@ -0,0 +1,11 @@
+import { Onboarding } from './StackExample'
```
7alip, line 1: I would give the component name the same as the filename to avoid confusion.

```suggestion
import { StackExample } from './StackExample'
```

7alip (commented):
Thanks, looks very clean!

Issue #39, opened 2022-08-07
Admin | PageHeader


7alip:
Design: https://www.figma.com/file/2KeHIr9IOh0HoCO89akiLn/WsvvUI?node-id=833%3A11557

This component is considered to be a reusable header component for some table and grid pages.

![Image](https://user-images.githubusercontent.com/22167684/183307366-9e38c75c-694c-42be-91a6-c27433dc6b7a.png)

Component has the following parts:
  - Title
  - Search bar
  - Filter menu button
  - Sort menu button
  - Action menu buttons

```ts
type PageHeaderProps = {
  title: string
  onSearch?: (value: string) => void
  filters?: Array<{ label: string; onClick: () => void }>
  sorts?: Array<{ label: string; onClick: () => void }>
  buttons?: Array<{ label: string; onClick: () => void }>
}
```

**Tasks:**

- [ ] Show component parts according to given props.
  - For example, show the search bar only if there is `onSearch` prop.
- [ ] Show menu list for `filter` and `sort` buttons. List array should be given as props. You can see an example filters prop for news page:

```ts
filters = [
  {
    label: "Bold",
    onClick: () => alert("Filter Bold news")
  },
   {
    label: "Nos",
    onClick: () => alert("Filter Nos news")
  }
]
```

PR #48, opened 2022-08-12
Admin PageHeader

yhtyyar94:


MustafaTRKYLMZ:
I think pull request should open from current branch to **dev**

7alip:
> I think pull request should open from current branch to **dev** . That is why there is lots of problem. You should reopen this pr **from Admin-PageHeader to dev**

I've changed it to dev. You can always edit to change the target branch.

Screenshot 2022-08-12 at 16 56 26

libs/ui/src/Admin/PageHeader.tsx
```diff
@@ -0,0 +1,166 @@
...
+  MenuItem,
+} from '@chakra-ui/react'
+import { HiOutlineFilter } from 'react-icons/hi'
+import { VscListFilter } from 'react-icons/vsc'
+
+interface PageHeaderProps {
+  onSearch: (item: string) => void
+  title: string
+  images: { slug: string; url: string }[]
+  textColor: string
```
7alip, line 22: We don't need text color here. Each project will have it's own `primary` color. Default primary color is `blue` from chakra. Just put the primary color to the header text.

libs/ui/src/Admin/PageHeader.tsx
```diff
@@ -0,0 +1,166 @@
...
+  backgroundColor: string
```
7alip, line 23: You can remove backgroundColor as well. Let's keep it transparent by default.

libs/ui/src/Admin/PageHeader.tsx
```diff
@@ -0,0 +1,166 @@
...
+  createNew: () => void
```
7alip, line 24: I think you should have added `buttons` `filters` and `sorts` instead of single action. 

Explained in #39 

```tsx
type PageHeaderProps = {
  title: string
  onSearch?: (value: string) => void
  filterMenu?: MenuItemProps[] // or ComponentProps<typeof MenuItem>[]
  sortMenu?: MenuItemProps[]
  buttons?: ButtonGroups[] // or ComponentProps<typeof Button>[]
}
```

libs/ui/src/Admin/PageHeader.tsx
```diff
@@ -0,0 +1,166 @@
...
+  return (
+    <Flex
+      justifyContent={{ base: 'space-between' }}
+      alignItems={'center'}
+      padding={1}
+      h={{ base: '60px' }}
+      backgroundColor={backgroundColor}
+      borderRadius={{ base: '10px' }}
+    >
+      <Text color={textColor} w={{ base: '40%' }} fontWeight={'bold'}>
```
7alip, line 48: Width is needed here?: 

Also, `base` applies from 0px bearkpoint if you don't give any other breakpoints. It's probably the same think as `w="40%"`. You can read an explanation about responsive styles here: https://github.com/wsvvrijheid/monorepo/pull/47#discussion_r944140256

```suggestion
      <Text color="primary.400"  fontWeight={'bold'}>
```

libs/ui/src/Admin/PageHeader.tsx
```diff
@@ -0,0 +1,166 @@
...
+            focusBorderColor={'none'}
+            value={search}
+            onChange={(e: any) => setSearch(e.target?.value)}
+            onKeyDown={(e: any) =>
+              e.key === 'Enter' ? onSearch(search) : null
+            }
+            onBlur={() => setFocusColor('lightgray')}
+            onFocus={() => setFocusColor(textColor)}
+          />
+        </Flex>
```
7alip, line 76: https://chakra-ui.com/docs/components/input/usage#add-elements-inside-input

You could use InputGroup, InputLeftElement for the icon to get more consistent design. You wouldn't need to give onFocus onBlur effects in this case.

We mostly use default Chakra component. We can later update our component styles in theme, so that all the components will have the same design.

libs/ui/src/Admin/PageHeader.tsx
```diff
@@ -0,0 +1,166 @@
...
+        <Menu>
+          <MenuButton>
+            <Avatar
+              src={
+                images.filter(image => Object.values(image)[0] === flags)[0].url
+              }
+              h={{ base: '40px' }}
+              w={{ base: '40px' }}
+              borderRadius={'50%'}
```
7alip, line 85:
```suggestion
              size="lg" // 
              rounded="full"
```

libs/ui/src/Admin/PageHeader.tsx
```diff
@@ -0,0 +1,166 @@
...
+          borderRadius={'10px'}
+        >
+          <Menu>
+            <MenuButton>
+              <HiOutlineFilter />
+            </MenuButton>
+            <MenuList mt={2} ml={-3}>
+              <MenuItem>Filter by Date</MenuItem>
+              <MenuItem>Filter by Author</MenuItem>
+              <MenuItem>Filter by Published</MenuItem>
```
7alip, line 135: We should be able to set filtering options from props. Each page can have different filtering options and functions. That's why we want to create a generic Header.

You can apply the same approach for sortMenu and buttons

```tsx
{filterMenu.map(filterItem => 
  <MenuItem {...filterItem} />
}
```

libs/ui/src/Admin/PageHeader.tsx
```diff
@@ -0,0 +1,166 @@
...
+          </Menu>
+        </Flex>
+        <Button
+          backgroundColor={textColor}
+          color={'white'}
+          marginLeft={5}
+          onClick={() => createNew()}
+        >
+          Create {title} +
+        </Button>
```
7alip, line 162: You can render buttons here.

```tsx
{buttons && 
  <ButtonGroup>
    {buttons.map(button => <Button {...button} />)}
  </ButtonGroup>
}
```

7alip (changes requested):
Thank you, nice job! 

Can you try to apply the changes and test all the functions like `alert("Filter clicked)`?

7alip:
Hey @yhtyyar94, hope you don't mind that I refactored your code to make the component more generic.

Could you review it (https://github.com/wsvvrijheid/monorepo/pull/48/commits/eec8ca23e5b901b46b5a560cb715968b7c382a36) and then merge if everything is ok?

PR #60, opened 2022-08-19
Feat/example chakra react query akin suleyman

rakin-tanis:


libs/ui/src/admin/MessageBox/MessageBox.stories.tsx
```diff
@@ -0,0 +1,51 @@
+import { Container } from '@chakra-ui/react'
+import { Meta, Story } from '@storybook/react'
+
+import { MessageBox } from './MessageBox'
+import '@splidejs/splide/dist/css/themes/splide-default.min.css'
```
7alip, line 5: You probably don't need it?

libs/ui/src/admin/MessageBox/MessageBox.stories.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { MessageBoxFieldValues, MessageBoxProps } from './types'
+
+export default {
+  component: MessageBox,
+  title: 'Admin/MessageBox',
```
7alip, line 10: As these components are just for onboarding purposes, I would rename it as `Example/MessageBox` so that we can easily differentiate them in storybook sidebar.

libs/ui/src/admin/MessageBox/MessageBox.tsx
```diff
@@ -0,0 +1,99 @@
+import React from 'react'
+
+import { Flex, useBreakpointValue, VStack } from '@chakra-ui/react'
+import { Textarea } from '@chakra-ui/react'
+import { Button } from '@chakra-ui/react'
```
7alip, line 5: You import them from one source (`@chakra-ui/react`), they could be imported at once!

libs/ui/src/admin/MessageBox/MessageBox.tsx
```diff
@@ -0,0 +1,99 @@
...
+      <Flex
+        minWidth="max-content"
+        justifyContent={{ base: 'center', lg: 'end' }}
+        gap="2"
+      >
+        {/* <ButtonGroup gap="2" w={{ base: 'full' }}> */}
+        <Button
+          size={size}
+          variant="solid"
+          colorScheme="gray"
```
7alip, line 64: They are probably default values, you might not need to give.

7alip (changes requested):
Hey @rakin-tanis @slymny, excellent! Thank you.

I would approve it but it's better to do the followings before merge:

1. I consider to keep these example components in `examples` folder. Could you move them (`MessageBox and Notification`) there?
2. ~I had already created an empty component file named `ChakraReactQueryExample.tsx`. I expected you to work in this file. **It's not required** but you can move your code into this file.~ And it's better to review the stories. Hopefully you can understand what's going on especially in the `useMutation` part.
3. I remember I moved `StackExample` to `examples` folder. You might have created your branch from an earlier version of dev. Could you delete this folder from `admin`?

___

**UPDATE:**

I forgot that multiple developers are working on the same issue. You don't need to apply the 2nd request. You can just move your files into `examples`

7alip (approved):
Looks perfect to me, thanks!
You can merge it now

PR #66, opened 2022-08-21
add cardbase component

ertugrul-aydin:


MustafaTRKYLMZ:
I think we can add hasZoom(true) prop in artprops. @7alip According to you?

7alip (approved):
Thank you

PR #84, opened 2022-08-23
publish and unpublish files converted to typescript

yhtyyar94:


libs/utils/src/services/art/publish.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export const publishArt = ({ id }: { id: number }) =>
+  mutation<Art>().put('api/arts', id, { data: { publishedAt: new Date() } })
+
+export const usePublishArt = (queryKey: string) => {
+  const queryClient = useQueryClient()
+  const toast = useToast()
+
+  return useMutation({
+    mutationKey: 'delete-art',
```
7alip, line 15: Mutation key can be `['publish-art', id]`

libs/utils/src/services/art/unpublish.ts
```diff
@@ -0,0 +1,47 @@
...
+
+export const unpublishArt = ({ id }: { id: number }) =>
+  mutation<Art>().put('api/arts', id, { data: { publishedAt: null } })
+
+export const useUnpublishArt = (queryKey: string) => {
+  const queryClient = useQueryClient()
+  const toast = useToast()
+
+  return useMutation({
+    mutationKey: 'delete-art',
```
7alip, line 15: Mutation key can be ['unpublish-art', id]

7alip (changes requested):
Nice work, you can just change the mutation keys.

PR #86, opened 2022-08-23
Feat/move contact

MustafaTRKYLMZ:
I added contact page


7alip (changes requested):
Thank you

7alip:
@MustafaTRKYLMZ I updated your code. Added useMutation for form submission. Created email service function.

You can review the changes here: https://github.com/wsvvrijheid/monorepo/pull/86/commits/2ddc4a45b5e989425a5ff20f5b84d3f16481cd6e

I think we shouldn't pass `errorMessage` coming from backend. Backend will send the errors in English and it can mostly be for developers. For now we can show only static error message. We can maybe later create an error handler to convert possible backend errors to user friendly translated error messages.

PR #87, opened 2022-08-23
Art find service

yhtyyar94:


libs/utils/src/services/art/find.ts
```diff
@@ -0,0 +1,90 @@
...
+      }
+
+  const categoryObj = qs.parse(categories)
+
+  const filters: { [key: string]: unknown } = {
+    ...(searchFilter || {}),
+    categories: { code: { $in: Object.values(categoryObj) } },
+    ...(statusFilter || {}),
+  }
+  return request()<Art>({
```
7alip, line 63:
```suggestion
  return request()<Art[]>({
```

We expect that `api/arts` query returns array of arts.

7alip (changes requested):
There is only one small typing issue, the rest seems correct to me. Thanks!

libs/utils/src/services/art/find.ts
7alip: Also don't forget to export it from index.ts

PR #88, opened 2022-08-23
terms

yhtyyar94:


7alip (changes requested):
After changing the imports you can test it by running `yarn start`. http://localhost:4200/terms works you can merge it.

apps/foundation/pages/terms.tsx
```diff
@@ -0,0 +1,56 @@
...
+    title: data.data?.title,
+    description: truncateText(data.data?.content || '', 200),
+  }
+
+  return {
+    props: {
+      term: data.data,
+      source,
+      seo,
+      ...(await serverSideTranslations(locale, ['common'])),
```
7alip, line 53: The solution that I found to make i18next work with nx monorepo is that you need to pass i18nConfig in every page when you use serverSideTranslations. You can check index or platform pages.

```suggestion
      ...(await serverSideTranslations(locale, ['common']), i18nConfig),
```

7alip (changes requested):

PR #91, opened 2022-08-26
join page

rakin-tanis:


libs/ui/src/templates/JoinTemplate/JoinTemplate.tsx
```diff
@@ -0,0 +1,132 @@
...
+import { mutation, request } from '../../../../utils/src/lib'
+import { Platform } from 'libs/types/src/platform'
+import { Job } from 'libs/types/src/job'
+import { PlatformList } from '../../components/PlatformList'
+import { JoinTemplateProps } from './types'
+import { useRouter } from 'next/router'
+import { StrapiLocale, Volunteer } from '@wsvvrijheid/types'
+import { JoinFormFieldValues } from '../../components/JoinForm/types'
+
+import { toastMessage } from '../../../../utils/src/util/toast'
```
7alip, line 27:
```suggestion
import { toastMessage } from '@wsvvrijheid/utils'
```

libs/ui/src/templates/JoinTemplate/JoinTemplate.tsx
```diff
@@ -0,0 +1,132 @@
...
+  Center,
+  SimpleGrid,
+  VStack,
+} from '@chakra-ui/react'
+import { useTranslation } from 'next-i18next'
+import { useMutation, useQuery } from 'react-query'
+import { v4 as uuidV4 } from 'uuid'
+
+import { Container, JoinForm, PageTitle } from '../../components'
+import { mutation, request } from '../../../../utils/src/lib'
```
7alip, line 18:
```suggestion
import { mutation, request } from '@wsvvrijheid/utils'
```

libs/utils/src/util/toast.ts
```diff
@@ -0,0 +1,30 @@
+import { AlertStatus, createStandaloneToast } from '@chakra-ui/react'
+
+// FIXME Test fails if we import with alias
+import { theme } from '../../../../libs/config/src/theme/theme'
+
+const { toast } = createStandaloneToast({
+  theme,
+})
```
7alip, line 8: Do we have to give `theme` here?

https://chakra-ui.com/getting-started/migration#2-update-createstandalonetoast

Also keep in mind that we need to add `<ToastContainer />` in `preview.tsx` to make the toast work.

7alip (changes requested):

libs/utils/src/util/toast.ts
rakin-tanis: I am not sure about `theme`, I copied the code directly from wsvvriheid. It is not mandatory giving params to `createStandaloneToast`. I can remove it.

There is a `_app.tsx` file in `/apps/foundation/pages/_app.ts` and it contains `<ToastContainer />`, I suppose you mean it.

libs/ui/src/templates/JoinTemplate/JoinTemplate.stories.tsx
```diff
@@ -0,0 +1,20 @@
+import { Story, Meta } from '@storybook/react'
+
+import { JoinTemplate, JoinTemplateProps } from '.'
```
7alip, line 3: I'm not sure if it causes circular dependency error but I would import them from their own files like following:

```suggestion
import { JoinTemplate } from './JoinTemplate'
import { JoinTemplateProps } from './types'
```

7alip: And if you don't have multiple component files in the folder, you don't need to create `types.ts`, you can put the type in the component file.

Issue #92, opened 2022-08-27
Update react-query to @tanstack/react-query


7alip:
React query announced v4. It's better to keep it up to date.

Migration notes https://tanstack.com/query/v4/docs/guides/migrating-to-react-query-4

PR #106, opened 2022-08-29
Update dependencies

7alip:
If everything works as expected we can merge it.

Related to #92 

We didn't notice that we merged wrong yarn.lock file generated by npm instead of yarn.


MustafaTRKYLMZ (approved):
Thank you for your effort

PR #118, opened 2022-09-01
Collection Page

yhtyyar94:


apps/foundation/pages/club/collection/[slug].tsx
```diff
@@ -0,0 +1,99 @@
+import { useEffect, useState } from 'react'
+import { useRef } from 'react'
+
+import { Center, Spinner, useBreakpointValue } from '@chakra-ui/react'
+import { Collection } from '@wsvvrijheid/types'
+import { Container, Layout } from '@wsvvrijheid/ui'
```
Comment (7alip) [inline apps/foundation/pages/club/collection/[slug].tsx:6]: Layout should be imported from foundation components not from our component library

7alip (changes requested):

apps/foundation/pages/club/collection/[slug].tsx
```diff
@@ -0,0 +1,91 @@
...
+
+  const title = collection.title || null
+
+  const seo = {
+    title,
+  }
+
+  return {
+    props: {
+      ...(await serverSideTranslations(locale, ['common'])),
```
Comment (7alip) [inline apps/foundation/pages/club/collection/[slug].tsx:85]: You forgot to provide next18nConfig. Building fails because of that. You can see other pages.

Comment (yhtyyar94) [reply apps/foundation/pages/club/collection/[slug].tsx:85]: ok i will fix it asap

PR #119, opened 2022-09-02
Api auth serverless pages

slymny:


libs/utils/src/services/auth/getAuth.ts
```diff
@@ -0,0 +1,43 @@
...
+import axios from 'axios'
+
+export const getAuth: (
+  identifier: string,
+  password: string,
+) => Promise<Auth | undefined> = async (
+  identifier: string,
+  password: string,
+) => {
+  const response = await axios.post(`${API_URL}/api/auth/local`, {
```
7alip, line 13: We can maybe create a AuthResponse type for login response in `libs/types/src/auth.ts`

```ts
type AuthResponse = {
  jwt: string;
  user: {
    id: number;
    username: string;
    email: string;
    provider: string;
    confirmed: boolean;
    blocked: boolean;
    createdAt: string;
    updatedAt: string;
  };
};
```

```suggestion
  const response = await axios.post<AuthResponse>(`${API_URL}/...
```

libs/utils/src/services/auth/getAuth.ts
```diff
@@ -0,0 +1,43 @@
...
+    identifier,
+    password,
+  })
+
+  const token = response.data.jwt
+  const userId = response.data.user?.ids
```
7alip, line 19: Should be `id`

```suggestion
  const userId = response.data.user?.id
```

apps/admin/pages/api/auth/social-auth/[provider].ts
```diff
@@ -0,0 +1,73 @@
...
+        url = `/api/auth/${provider}/callback?access_token=${req.body.access_token}&access_secret=${req.body.access_secret}`
+      }
+
+      const socialLoginResponse = await fetcher(req.body.access_token).get(url)
+
+      const token = socialLoginResponse.data.jwt
+      const userId = socialLoginResponse.data.user?.id
+
+      if (token) {
+        const artistResponse = await request(token)({
```
Comment (7alip) [inline apps/admin/pages/api/auth/social-auth/[provider].ts:22]:
```suggestion
        const artistResponse = await request(token)<Artist[]>({
```

apps/admin/pages/api/auth/social-auth/[provider].ts
```diff
@@ -0,0 +1,73 @@
...
+              user: userId,
+              name: socialLoginResponse.data.user?.username,
+            },
+          })
+        }
+
+        const response = await request(token)({
+          url: 'api/users',
+          filters: { id: { $eq: userId } },
+        })
```
Comment (7alip) [inline apps/admin/pages/api/auth/social-auth/[provider].ts:39]: Probably it makes more sense to fetch the user by id.

```suggestion
        const response = await request(token)<User>({
          url: `api/users/${userId}`,
        })
```

apps/admin/pages/api/auth/reset-password.ts
```diff
@@ -0,0 +1,52 @@
...
+        `${process.env.NEXT_PUBLIC_API_URL}/api/users/me?populate=*`,
+        {
+          headers: { Authorization: `Bearer ${token}` },
+        },
+      )
+
+      const auth = { user: response.data, token, isLoggedIn: true }
+
+      // req.session as Session
+      ;(req.session as Session).user = auth.user
```
7alip, line 34: This line seems unnecessary since we already pass auth object in the following line

apps/admin/pages/api/auth/register.ts
```diff
@@ -0,0 +1,48 @@
...
+    const token = response.data.jwt
+    const userId = response.data.user?.id
+
+    await mutation(token).post('api/artists', {
+      data: { user: userId, name: trimmedName },
+    })
+
+    const auth = await getAuth(email, password)
+
+    req.session = auth as Session
```
7alip, line 33: It's safer to keep session props like `destroy` and `save`

```suggestion
    req.session = { ...auth, ...req.session }
```

apps/admin/pages/api/auth/social-auth/[provider].ts
```diff
@@ -0,0 +1,73 @@
...
+            id: userData.id,
+            username: userData.username,
+            email: userData.email,
+            volunteer: userData.volunteer,
+            avatar: userData.avatar,
+            artist: userData.artist,
+          }
+
+          const auth = { user, token, isLoggedIn: true }
+          req.session = auth as Session
```
Comment (7alip) [inline apps/admin/pages/api/auth/social-auth/[provider].ts:54]: Again, it's safer to keep session props like `destroy` and `save` like you did in reset-password.

```suggestion
    req.session = { ...auth, ...req.session }
```

apps/admin/pages/api/auth/login.ts
```diff
@@ -1,45 +1,16 @@
...
-      }
+    const auth = await getAuth(identifier, password)
 
-      ;(req.session as Session).user = user
-      ;(req.session as Session).isLoggedIn = true
-      ;(req.session as Session).token = token
-      await req.session.save()
-      res.json({ user, isLoggedIn: true, token })
-    }
+    req.session = auth as Session
```
7alip, line 11: It's safer to keep session props like `destroy` and `save`

```suggestion
    req.session = { ...auth, ...req.session }
```

7alip (changes requested):
Thank you. There are some small improvements.

Before merging it, I'll test them in real pages.

libs/utils/src/services/auth/getAuth.ts
slymny: There are also artist, avatar, volunteer props in the response. Instead of adding this should I use existing User type and add non-existing props?
New suggestion: 
type AuthResponse = {
  jwt: string
  user?: User | null
  id: number | null
  createdAt: string | null
  updatedAt: string | null
}

apps/admin/pages/api/auth/register.ts
slymny: It's for casting. There is no problem with the other properties of the auth object, but only for the user. Is it safe to not cast at all? Because otherwise there is no casting.

libs/utils/src/services/auth/getAuth.ts
7alip: I'm not sure if the `post` request returns extended user data. We might need to adjust the backend for `/api/auth/local` request. Populated user data returns for the request `/api/users`

apps/admin/pages/api/auth/register.ts
7alip: I remember req.session object has `save` and `destroy` property functions. I'm just curious if we re assign the session object, will we loose the function properties.

```ts
session = {
  save: () => void
  destroy => void
}
```

Then we try to reassign the session like:

```ts
session = {
  user: User
  token: string
  isLoggedIn: boolean
}
```

That's why I had to do the trick `auth as Session` because typescript was complaining about `save` and `destroy` props which are required props of the session from the library.

We could also do

session.auth = auth or (then adjust the Session type)
session.user = auth.user, session.token = auth.token, session.isLoggedIn = true

PR #121, opened 2022-09-04
Session User

7alip:
I haven't planned how we will control user roles, pages, permissions etc. This PR is a simple configuration for the session. Depending on our needs, we may want to extend it later.

Since session cookie has limitations ([see discussion](https://github.com/vvo/iron-session/issues/375)), I need to simplify the session data. Probably we will show/hide menus and fetch page related data based on this session.

Could you review and if it's ok for you, you can merge it into your branch #119 and then merge your branch into dev.


slymny (approved):
Looks good

PR #122, opened 2022-09-04
blog dynamic page

zahirekrem09:
#81


apps/foundation/pages/blog/[slug].tsx
```diff
@@ -1,17 +1,124 @@
...
+      acc[l.locale] = l.slug
+      return acc
+    }, {}) || {}
+
+  const title = blog?.title || null
+  const description = blog?.description || null
+  const adminUrl = API_URL
+  const siteUrl = SITE_URL
+  const image = blog.image
+  const url = `${siteUrl}/${locale}/blog/${slugs[locale]}`
```
Comment (7alip) [inline apps/foundation/pages/blog/[slug].tsx:79]: I think slugs doesn't contain the current blog's slug but it's localizations. We use slugs from pageProps when we switch the language.

```suggestion
  const url = `${siteUrl}/${locale}/blog/${blog.slug}`
```

apps/foundation/pages/blog/[slug].tsx
```diff
@@ -1,17 +1,124 @@
...
+              secureUrl: adminUrl + image?.url,
+              type: image?.mime,
+              width: image?.width,
+              height: image?.height,
+              alt: title,
+            },
+          ]
+        : [],
+    },
+  }
```
Comment (7alip) [inline apps/foundation/pages/blog/[slug].tsx:110]: Can you move it from getBlogStaticProps service. We may use the same page in some of other sites. _(Storyblok may fail because of some server side works, like serialize and serverSideTranslations. That's why I don't expect them in getBlogStaticProps function. Once we do module imports in a proper way, we can move all the logic later)_

```ts
const { blog, authorBlogs, seo, queryClient } = await getBlogStaticProps()

if(!blog) return { notFound: true }

const source = await serialize(blog.content || '')

return {...}
```

PR #128, opened 2022-09-06
Admin Layout

7alip:
Could you merge it if everything works


MustafaTRKYLMZ (approved):
Everything looks good. Only there is problem some page because of we don't have table in the backend. For example feedbacks/foundation.

PR #130, opened 2022-09-07
Admin Collapsible Navbar

7alip:
- Adds collapse/expand button at the bottom of the sidebar
- Changes default primary color as green
- Changes default font as Poppins
- Fixes loading indicator for ssg pages

### DIFF
--- a/apps/admin/pages/_document.tsx
+++ b/apps/admin/pages/_document.tsx
@@ -28,7 +28,7 @@ export default class Document extends NextDocument {
           <link rel="preconnect" href="https://fonts.gstatic.com" />
           <link rel="preconnect" href="https://fonts.googleapis.com" />
           <link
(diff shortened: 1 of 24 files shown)


MustafaTRKYLMZ:
Screenshot_20220907-084645_Chrome

In the monbil version  difficult to see the collops. Maybe we can fix navbar vertical size

7alip:
> In the monbil version difficult to see the collops. Maybe we can fix navbar vertical size

I didn't consider collapsed design for mobile actually. The idea is to leave more space for the user even on desktop.

For mobile users we can add menu icon that opens the nav inside a drawer. But we can try to fix the layout issue you mentioned in either way.

MustafaTRKYLMZ (approved):

PR #135, opened 2022-09-09
Feat/club page

rakin-tanis:


7alip:
@rakin-tanis Thanks for amazing work! 👏

I'll review it soon, but the only problem I noticed is, searchTerm is not always being cleaned from the url probably because of debouncing time. You can maybe check it if you find a time

libs/ui/src/templates/ArtClubTemplate/ArtClubTemplate.tsx
```diff
@@ -0,0 +1,189 @@
...
+  const queryKey = ['arts', locale, searchTerm, categories || null, page || '1']
+
+  // Custom useQuery hook or fetching arts
+  const artsQuery = useArts(queryKey, {
+    categories: categories as string,
+    populate: ['artist.user.avatar', 'categories', 'images', 'likers'],
+    page: parseInt(page as string) || 1,
+    pageSize: 12,
+    searchTerm: searchTerm as string,
+    username: auth.user?.username,
```
7alip, line 64: I don't think we want to search the arts by current authenticated user's usernaname.

```ts
const userFilter = {
    artist: {
      user: {
        username: {
          $containsi: searchTerm || username,
        },
      },
    },
  }
  ```

  As you can see in `userFilter`, searchTerm is filters for both username and art title. Do you think we should use `username` parameter in this case?

File: libs/ui/src/templates/ArtClubTemplate/ArtClubTemplate.tsx [hunk 2]
```diff
@@ -0,0 +1,189 @@
...
+    sort: ['publishedAt:desc'],
+    locale: locale as StrapiLocale,
+  })
+
+  useUpdateEffect(() => {
+    artsQuery.refetch()
+  }, [])
```
7alip, line 71: Dependency array is empty. In which case it will refetch then?

I think, useArts has dynamic parameters. If I'm not mistaken, once any of these parameters changes, re-rendering will trigger refetching.

libs/utils/src/services/art/like.ts
```diff
@@ -60,10 +60,11 @@ export const useLikeArt = (
...
   const isLikedByUser =
-    user &&
-    art.likers &&
-    art.likers?.length > 0 &&
-    art.likers?.some(({ id }) => id === user.id)
+    (user &&
+      art.likers &&
+      art.likers?.length > 0 &&
+      art.likers?.some(({ id }) => id === user.id)) ||
+    undefined
```
7alip, line 67: I don't remember why we wrote like this but wouldn't it also work like this:

`user && art.likers?.some(({ id }) => id === user.id))`

But let's skip this for now. We need to check `isLikedByUser` from the backend.

libs/ui/src/components/ArtCard/ArtCardBase.tsx
```diff
@@ -15,20 +15,30 @@ import { API_URL } from '@wsvvrijheid/config'
...
+export const ArtCardBase: FC<ArtCardBaseProps> = ({
   art,
   isMasonry,
   toggleLike,
   isLiked,
   actions,
   isOwner,
+  isModal = false,
 }) => {
+  const auth = useAuth()
```
7alip, line 35: We probably didn't noticed some points in the old code. I need to point out something about useAuth here:

`useAuth()` can make an api call each time we use it to check authentication. Until we find a better approach, I wouldn't recommend useAuth to be used in components, especially the components that are being rendered in a list, like Cards. I try to use it in page components and pass to the children.

But here, if you follow `auth`, we use it only in ArtWithDetails. Wouldn't it be better to use useAuth there?

My recommendation is:

- Move ArtModal into ArtCard. 
- Remove isModal prop. Pass href from ArtCard (href from ArtCardBaseProps)
- You don't need to change Navigate in this case. 
- Pass auth object to ArtModal from ArtCard.

7alip (changes requested):
Thank you for your great efforts.
I noticed some points can be improved.

libs/ui/src/templates/ArtClubTemplate/ArtClubTemplate.tsx
rakin-tanis: you're right, I'll change it.

libs/ui/src/templates/ArtClubTemplate/ArtClubTemplate.tsx
rakin-tanis: There was only the `searchTerm` parameter in the old code's dependency array of `useUpdateEffect`. After I added `searchTerm` as a `query parameter`, I removed it from the `useUpdateEffect` dependencies. Because as I know, every time `searchTerm` is changed, it is pushed as a query parameter and then the page is rendered again and `useArts` runs one time. I think you are partially right, we do not need `useUpdateEffect` anymore.

We send those parameters to the `useArts` hook:
**categories:  comes from url**
**page:  comes from url**
**searchTerm:  comes from url**
**locale:  comes from url**
populate: not change
pageSize: not change
sort: not change

The first four of them are pushed as the query parameters so, it causes the page to be rendered again, and it triggers `useArts` to run.

libs/ui/src/components/ArtCard/ArtCardBase.tsx
```diff
@@ -126,8 +136,22 @@ export const ArtCardBase: FC<ArtCardProps> = ({
...
+                aria-label="view art"
+                borderColor="whiteAlpha.500"
+                borderWidth={1}
+                color="white"
+                colorScheme="blackAlpha"
+                icon={<FaExternalLinkSquareAlt />}
+                rounded="full"
+              />
+            </Navigate>
+          )}
```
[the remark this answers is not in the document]
rakin-tanis: it looks cleaner.

libs/ui/src/hooks/useChangeParam.tsx
```diff
@@ -1,46 +1,68 @@
...
+const cleanQuery = (query: ParsedUrlQuery, args: ChangeParamArgs) => {
+  const newQuery = Object.assign({}, query)
+  for (const key in args) {
+    const param = args[key]
+    if (
+      param === '' ||
+      param === null ||
+      param === undefined ||
+      param.length === 0
+    ) {
```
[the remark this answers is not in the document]
rakin-tanis: if we change like that we can't send `0` or `true`, or `false` values as a query parameter.

libs/utils/src/services/art/like.ts
rakin-tanis: you're right, checking the length of the array is redundant here.

7alip (approved):
I really appreciate your work 💯.

There were some conflicts with dev and I had to fix them. I also added `keepPreviousData` and loading for searchbar.

You can merge it if you want to.

PR #147, opened 2022-09-15
Feat/create art mutation

rakin-tanis:


apps/foundation/pages/club/index.tsx
```diff
@@ -28,8 +28,15 @@ export const getStaticProps: GetStaticProps = async context => {
     // So, we need to keep the same order of the `queryKey` array
 
     // queryKey: [arts, locale, searchTerm, category, page]
-    queryKey: ['arts', locale, '', null, '1'],
-    queryFn: () => getArts({ locale: locale as StrapiLocale }),
+    queryKey: ['arts', locale, null, null, '1'],
+    queryFn: () =>
+      getArts({
+        populate: ['artist.user.avatar', 'categories', 'images', 'likers'],
```
7alip, line 34: As you might have already noticed, Art-Artist relation is now directly related to the User

```suggestion
        populate: ['artist.avatar', 'categories', 'images', 'likers'],
```

libs/ui/src/templates/ArtClubTemplate/ArtClubTemplate.tsx
```diff
@@ -64,8 +75,40 @@ export const ArtClubTemplate: FC = () => {
...
+
+    const slug = slugify(data.title)
+    const formBody = {
+      ...data,
+      slug,
+      artist: auth.user.id,
+      categories: data.categories?.map(c => Number(c.value)),
+    }
+
+    mutate(formBody)
```
7alip, line 111: For an alternative solution, you can also use useMutation callbacks.

```suggestion
    mutate(formBody, {
      onSuccess: createArtOnSuccess,
      onError: createArtOnError
    })
```

libs/ui/src/templates/ArtTemplate/ArtTemplate.tsx
```diff
@@ -27,7 +27,9 @@ export const ArtTemplate: FC<ArtTemplateProps> = ({ auth, queryKey }) => {
 
   useViewArtMutation()
 
-  const categories = (art?.categories?.flatMap(c => c.slug) || []) as string[]
+  const categories = (art?.categories?.flatMap(
+    (c: { slug: string }) => c.slug,
```
7alip, line 31: Is there a reason for definin type here?

libs/utils/src/services/art/create.ts
```diff
@@ -0,0 +1,22 @@
...
+export const createArt = (artCreateInput: ArtCreateInput) => {
+  return createMutation<Art, ArtCreateInput>('api/arts', artCreateInput)
+}
+
+export const useCreateArt = ({
+  onSuccess,
+  onError,
+}: {
+  onSuccess?: () => void
+  onError?: () => void
```
7alip, line 15: As mentioned before, you could also use useMutation callbacks

libs/ui/src/components/CreateArtForm/types.ts
```diff
@@ -13,9 +15,14 @@ export type CreateArtFormFieldValues = {
 
 export type CreateArtFormProps = {
   onCreateArt: (data: CreateArtFormFieldValues & { images: Blob[] }) => void
-  isLoading: boolean
   isLoggedIn: boolean
-  categories: Category[]
+  isLoading: boolean
+  cancelRef: RefObject<HTMLButtonElement>
```
7alip, line 20: We can add `isSuccess` coming from useMutation. So that we can trigger ArtSuccessAlert with that boolean.

Also `isError` or `error` could be given via props to show error message underneath the form?

7alip (commented):
Special thanks for improving previous code

libs/ui/src/templates/ArtTemplate/ArtTemplate.tsx
rakin-tanis: eslint was complaining.

libs/ui/src/components/ArtCard/ArtCardBase.tsx
```diff
@@ -39,6 +39,13 @@ export const ArtCardBase: FC<ArtCardBaseProps> = ({
   } = useDisclosure()
 
   const [actionType, setActionType] = useState<ArtActionType>()
+  const [hover, setHover] = useState({ color: 'gray.100' })
+  const [color, setColor] = useState('white')
+
+  useEffect(() => {
+    setHover({ color: isLiked ? 'red.200' : 'gray.100' })
+    setColor(isLiked ? 'red.400' : 'white')
+  }, [])
```
7alip, line 48: It seems the color is not being updated because you set the colors only at the first render. 
By the way, was the prior code not updating the color after a click? As far as I know,  As far as I know, a component needs to be re-rendered if the prop changes.

```suggestion
  }, [isLiked])
```

libs/ui/src/components/CreateArtForm/CreateArtForm.stories.tsx
```diff
@@ -1,55 +1,38 @@
...
+      id: 38,
+      username: 'Artist',
+      email: 'artist@samenvvv.nl',
+      provider: 'local',
+      confirmed: true,
+      isAuthenticated: true,
+    },
+    isLoggedIn: true,
+    token: 'fake-token',
+  },
```
7alip, line 37: You could use mock data as well. Also, the user data you provided is the result of the login request. We map the user data for the session. 

It might be safer to give type for the args as well.

```ts
Auth.args = {
  auth: {
    user: mapSessionUser(sample(USER_MOCKS)),
    isLoggedIn: true,
    token: 'fake-token',
  },
} as CreateArtFormProps
```

7alip (approved):
Thank you so much for your efforts.

You can merge it after fixing the color update problem.

PR #150, opened 2022-09-17
Feat/admin art list

MustafaTRKYLMZ:
added art table


7alip (changes requested):
WTable component is created. You can see some examples in storybook

PR #151, opened 2022-09-17
Feat/artist page

zahirekrem09:
#145


libs/utils/src/services/artist/getByUsername.ts
```diff
@@ -0,0 +1,47 @@
...
+    },
+    populate: [
+      'arts.images',
+      'arts.artist',
+      'arts.artist.avatar',
+      'arts.categories',
+      'arts.localizations',
+      'arts.comments.user.avatar',
+      'arts.likers',
+    ],
```
7alip, line 29: This looks very heavy to me. We need to optimize it. Maybe we can fetch comments in a separate query for an individual art in details component.

Let's do it later

7alip (approved):
Looks good, thank you 👍

libs/utils/src/services/artist/getByUsername.ts
7alip: @zahirekrem09 Do you want to merge it?

PR #153, opened 2022-09-18
Login page

slymny:


apps/foundation/pages/api/auth/register.ts
```diff
@@ -0,0 +1,47 @@
...
+  const trimmedUsername = username.trim()
+  const trimmedEmail = email.trim()
+
+  try {
+    const response = await axios.post(
+      `${process.env.NEXT_PUBLIC_API_URL}/api/auth/local/register`,
+      {
+        username: trimmedUsername,
+        email: trimmedEmail,
+        password,
```
7alip, line 19: I should have inform you that we update the User. It can also receive name while registering. It's better to provide `name` as well.

```ts
        name: trimmedName
        username: trimmedUsername,
        email: trimmedEmail,
        password,
```

apps/foundation/pages/api/auth/register.ts
```diff
@@ -0,0 +1,47 @@
...
+      },
+    )
+
+    const token = response.data.jwt
+    const userId = response.data.user?.id
+
+    const body = { user: userId, name: trimmedName }
+
+    await createMutation('api/users', body, token)
```
7alip, line 28: With this process, we had aimed to create Artist-User relation for every registered users. Since we remove Artist table from the database, it's no longer needed. You should remove that code.

Now, a User can also be considered as Artist.

7alip (changes requested):
Thank you for the submission, nice job!
We just still need some small changes.

7alip:
@slymny Do you want to merge it?

PR #163, opened 2022-09-21
added profile page

zahirekrem09:


libs/utils/src/services/art/getByArtist.ts
```diff
@@ -0,0 +1,37 @@
...
+    populate: [
+      'artist.avatar',
+      'categories',
+      'images',
+      'localizations',
+      'comments.user.avatar',
+      'likers',
+    ],
+    locale,
+  })
```
7alip, line 25: Unpublished (publishedAt: null) arts owned by the artist should be listed as well.

You need to add `publicationState: "preview"` to request function.

7alip (approved):
Excellent! I just noticed one problem, after fixing it, you can merge.

apps/foundation/pages/profile/index.tsx
```diff
@@ -0,0 +1,50 @@
...
+  req,
+  locale,
+}) {
+  const auth = req.session
+
+  if (!auth.user) {
+    return {
+      redirect: {
+        permanent: false,
+        destination: '/user/login',
```
7alip, line 33:
```suggestion
        destination: '/login',
```

PR #167, opened 2022-09-22
News card

yhtyyar94:
there are some errors in NewsCard.tsx


libs/ui/src/admin/NewsCard/NewsCard.tsx
```diff
@@ -0,0 +1,76 @@
...
+      } else {
+        parsedBookmarks = parsedBookmarks.splice(1, parsedBookmarks.indexOf(id))
+        localStorage.setItem('bookmarks', JSON.stringify(parsedBookmarks))
+        setIsBookmarked(false)
+      }
+    } else {
+      localStorage.setItem('bookmarks', JSON.stringify([id]))
+      setIsBookmarked(true)
+    }
+  }
```
7alip, line 36: It could be easier to do with `useLocalStorage` from `react-use`.

An example solution could be like this:

```tsx
const [bookmarksStorage, setBookmarksStorage] = useLocalStorage(
  'bookmarks',
  [],
)

const isBookmarked = bookmarksStorage.some(id => id === news.id)

const handleBookmark = () => {
  if (isBookmarked) {
    setBookmarksStorage(bookmarksStorage.filter(id => id !== news.id))
  } else {
    setBookmarksStorage([...bookmarksStorage, news.id])
  }
}
```

libs/ui/src/admin/NewsCard/NewsCard.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+  const handleShare = (id: string) => {}
+
+  const handleView = (url: string) => {
```
7alip, line 40: With this approach (CardBase => Card) you don't need to pass ids or urls as we can provide them from news (news.id, news.url)

libs/ui/src/admin/NewsCardBase/types.ts
```diff
@@ -0,0 +1,22 @@
...
+export type NewsCardBaseProps = {
+  hideDescription?: boolean
+  news: News
+  variant?: 'horizontal' | 'vertical'
+  isBookmarked: boolean
+  onBookmark: (id: string) => void
+  onShare: (id: string) => void
+  onRecommend: () => void
+  onView: (url: string) => void
+}
```
7alip, line 15: The reason I'm using `RecommendedTopic` instead of `RecommendedNews` is that I couldn't use `News` as a database table name because it was complaining about pluralization the names.

Maybe it's a better idea to use `Topics`, `TopicCardBase` and `TopicCard` to avoid the confusion.

```suggestion
import { Topic } from '@wsvvrijheid/types'

export type TopicCardBaseProps = {
  hideDescription?: boolean
  topic: Topic
  variant?: 'horizontal' | 'vertical'
  isBookmarked: boolean
  onBookmark: (id: string) => void
  onShare: (id: string) => void
  onRecommend: () => void
  onView: (url: string) => void
}
```

libs/ui/src/admin/NewsCard/NewsCard.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+export const NewsCard: FC<NewsCardProps> = ({ news, userId, ...rest }) => {
+  const [isBookmarked, setIsBookmarked] = useState(false)
+  const toast = useToast()
+  const recommendMutation = useMutation({
+    mutationKey: ['recommend-topic'],
+    mutationFn: () =>
+      createMutation<RecommendedTopic, RecommendedTopicCreateInput>(`api/recommended-topics`, 
+        { ...news, recommender: userId },
+      )
```
7alip, line 17: We have to provide what `RecommendedTopicCreateInput` expects. We can fix type issues like following but I recommend changing `News` (Topic) type to make it similar to `RecommendedTopic`.

```ts
mutationFn: () =>
  userId &&
  createMutation<RecommendedTopic, RecommendedTopicCreateInput>(
    `api/recommended-topics`,
    {
      date: news.date,
      description: news.description,
      title: news.title,
      recommender: userId,
      imageUrl: news.image,
      publisher: news.owner,
      link: news.url,
    }
  )
```

---

I noticed that `News` type has not the same fields as `RecommendedTopic`. We can change `News` (Topic) type as following:

```ts
type Topic = Pick<
  RecommendedTopic,
  "id" | "title" | "description" | "imageUrl" | "date" | "publisher" | "url"
>
```

We should of course adjust NewsCards (TopicCard) based on the new type

7alip (changes requested):
Thank you for the submission. Here is my review comments.

libs/ui/src/admin/NewsCard/NewsCard.tsx
yhtyyar94: i think url should be link because there is no type called URL in ReccomendTopic

7alip (approved):
Great! Thanks a lot. I replaced all news with topic. Do you want to merge it?

PR #198, opened 2022-09-27
Feat/create post form

MustafaTRKYLMZ:
create tweet form added


package.json
```diff
@@ -54,6 +54,8 @@
     "react-redux": "^8.0.2",
     "react-router-dom": "^6.3.0",
     "react-use": "^17.4.0",
+    "string-similarity": "^4.0.4",
+    "string-similarity-js": "^2.1.4",
```
7alip, line 58: `string-similarity` seems to be more popular than the other. You can choose one of them and remove the other.

libs/ui/src/components/CreateTweetForm/types.ts
```diff
@@ -0,0 +1,12 @@
+export type CreateTweetFormProps = {
+  title: string
+  image: string
+  onSubmitHandler: () => void
```
7alip, line 4: How will we get the form values from the component? We should be able to take the following props from the component:

`media, originalTweet, text`

> (recommender: number (userId)) can bi given from the session user data

```ts
// RecommendedTweet model
type RecommendedTweet = {
  recommender: number
  originalTweet: Tweet
  media: Blob
  text: string
}
```

libs/ui/src/components/CreateTweetForm/types.ts
```diff
@@ -0,0 +1,12 @@
...
+  onClose: () => void
+  isOpen: boolean
+  setImages: React.Dispatch<React.SetStateAction<Blob[]>>
+  maxSize?: number
+  images: Array<Blob>
+  tweetContent: string
```
7alip, line 10: We will get `tweet data (original data from Twitter)` from human rights tweet timeline lists. I recommend passing all tweetData to the modal form instead of `tweetContent`. Because we will save `original tweetData` to our db as well.

libs/ui/src/components/CreateTweetForm/types.ts
```diff
@@ -0,0 +1,12 @@
+export type CreateTweetFormProps = {
+  title: string
+  image: string
+  onSubmitHandler: () => void
+  onClose: () => void
+  isOpen: boolean
+  setImages: React.Dispatch<React.SetStateAction<Blob[]>>
+  maxSize?: number
```
7alip, line 8: You can remove `maxSize`. I think we don't need to allow users to set the image size.

libs/ui/src/components/CreateTweetForm/CreateTweetForm.tsx
```diff
@@ -0,0 +1,148 @@
...
+                  {tweetImage?.length > 1 ? (
+                    <WImage src={tweetImage} hasZoom={true} alt={''} />
+                  ) : (
+                    <>
+                      <Text color={'black'} fontWeight={'bold'} w={'full'}>
+                        Add Image(s)
+                      </Text>
+                      <FileUploader images={images} setImages={setImages} />
+                    </>
+                  )}
```
7alip, line 118: I'm not sure if I understand this logic.

1. Why we show tweetImage only if there are more than 1 image? What about if there is only 1 tweet image?
2. Even there is a original tweet image, shouldn't we allow user to upload and recommend a new image. I think original image will always be available, because it will be in original tweetData. 
3. Shouldn't we always allow the users to be able to upload a new image.

libs/ui/src/components/CreateTweetForm/CreateTweetForm.stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+  // const createRecomendedTweet = useRecomendedTweet(queryKey)
+
+  const handleSubmit = async (tweet: string) => {
+    const recomendedTweet: RecommendedTweet | RecommendedTweetCreateInput = {
+      recommender: USER_MOCKS?.[0].id,
+      originalTweet: args.originalTweet,
+      media: images,
+      text: tweet,
+      isArchived: false,
+      isShared: false,
```
7alip, line 39: We didn't think they were necessary for the create mutation.

7alip (changes requested):

libs/ui/src/components/CreateTweetForm/CreateTweetForm.stories.tsx
```diff
@@ -0,0 +1,72 @@
...
+      <Button onClick={() => handleSizeClick()} m={4}>
+        {`Create Tweet`}
+      </Button>
+      <CreateTweetForm
+        {...args}
+        onSubmit={handleSubmit}
+        isOpen={isOpen}
+        onClose={onClose}
+        images={images}
+        setImages={setImages}
```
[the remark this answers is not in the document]
MustafaTRKYLMZ: I think it is necessary. Because It is normal in the real twitter account. We should upload according to real twitter

libs/ui/package.json
```diff
@@ -1,4 +1,7 @@
 {
   "name": "@wsvvrijheid/ui",
-  "version": "0.0.1"
+  "version": "0.0.1",
+  "devDependencies": {
+    "@types/string-similarity": "^4.0.0"
```
7alip, line 5: This should be in root package.json

MustafaTRKYLMZ:
It is ok

PR #200, opened 2022-09-28
Ms kose

MSKose:


libs/ui/src/examples/Fetch/E3_FetchRequest.tsx
```diff
@@ -3,27 +3,27 @@ import { useEffect, useState } from 'react'
...
+  const [blogs, setBlogs] = useState<Blog[]>([])
 
   useEffect(() => {
     // TODO: fetch blogs with our custom request function
+
+    const getBlogs = async () => {
+      try {
+        // const { data } = await fetcher(API_TOKEN)(BLOG_URL)
+        const { data } = await request({ url: 'api/blogs' })
+        setBlogs(data) // I have no idea why ts is saying this is null, but I get my data on the screen just fine
```
7alip, line 22: This is because we specified in `request` function that sometimes the fetched data might be null. But it's not 100% correct, we need to improve request return types.

Strapi returns `[]` empty array or `null` depending on the relation.

A. When Strapi returns `null`?
  1. Example: `blog.title` is not be filled (own property)
  2. Example: `blog.author` is not be filled (single relational data)
  3. Example: `/api/blog/1` we expect single blog. If there is no blogs, it will return `null`

B. When Strapi return `[]` empty array?
  1. Example: `blog.categories` is not be filled (multiple relational data)
  2. Example `/api/blogs` we expect multiple data. If there is no blogs it will return `[]`

7alip (commented):
Thanks, looks great. You can continue the other tasks without merging this.

7alip:
@MSKose You can close this PR as it was just for learning purposes.
I'll add more onboarding examples and you can create a new branch after I announce it. 
It will be easier for me to review your next work with a new PR.

PR #203, opened 2022-09-30
Admin collection table

slymny:


apps/admin/pages/arts/collections/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+  })
+  const handleSearch = (search: string) => {
+    search ? setSearchTerm(search) : setSearchTerm(undefined)
+  }
+
+  useUpdateEffect(() => {
+    CollectionsQuery.refetch()
+  }, [locale, searchTerm, sort])
+
+  const Collections = CollectionsQuery?.data?.data
```
7alip, line 36: camelCase is preferable for variable names

```suggestion
  const collections = CollectionsQuery?.data?.data
```

libs/ui/src/admin/DataTables/CollectionsTable/CollectionsTable.tsx
```diff
@@ -0,0 +1,37 @@
...
+  data: collections,
+  user,
+  totalCount,
+  currentPage,
+  onSort,
+  setCurrentPage,
+}) => {
+  return (
+    <>
+      {user && <CreateCollectionModal />}
```
7alip, line 26: The page will be visible only for authenticated users. So, I think it's not necessary to use `user` here

apps/admin/pages/arts/collections/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+        onLanguageSwitch: locale => setLocale(locale),
+        defaultLocale,
+        sortMenu: [
+          <MenuItem key="asc" icon={<FaArrowUp />}>
+            Name Asc
+          </MenuItem>,
+          <MenuItem key="desc" icon={<FaArrowDown />}>
+            Name Desc
+          </MenuItem>,
+        ],
```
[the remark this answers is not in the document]
7alip: Yes, I had considered to use it before I came up with column base sorting solution.

We can still use header sort menu for not table but grid pages. Or it can be used for advanced sorting options later.

libs/ui/src/admin/CreateCollectionModal/CreateCollectionModal.tsx
```diff
@@ -59,14 +59,6 @@ export const CreateCollectionModal: FC<CreateCollectionModalProps> = ({
...
   })
 
-  useEffect(
-    () => () => {
-      // Make sure to revoke the data uris to avoid memory leaks
-      images.forEach(image => URL.revokeObjectURL((image as any).preview))
-    },
-    [images],
-  )
-
```
7alip, line 69: Removed useEffect as we now use previews inside the file picker

7alip (approved):
Thanks, looks good to me.
We will think about localization later. You can merge it if you want

PR #214, opened 2022-10-08
fixed delete mutation

zahirekrem09:


7alip (approved):
Thanks for the fix. Could you merge it?

PR #215, opened 2022-10-09
Ms kose

MSKose:


Selman-S (commented):
good job!

libs/ui/src/examples/Fetch/E6_RequestUseQuery.tsx
```diff
@@ -0,0 +1,42 @@
...
+  const instance = axios.create({
+    headers: { Authorization: 'Bearer ' + TOKEN },
+  })
+
+  const getBlogs = async () => {
+    return await instance.get(BLOG_URL)
+  }
+
+  const { data, status } = useQuery(['title'], getBlogs)
+  // TODO Add Input to change nameFilter
```
7alip, line 28: Although author name filtering would be great to have, I think the main goal of this PR is to use `useQuery`, therefore it seems pretty good to me, thanks :)

7alip (approved):

PR #220, opened 2022-10-15
Feat/topic page

rakin-tanis:


apps/admin/pages/content-maker/news.tsx
```diff
@@ -1,12 +1,156 @@
...
+        </MenuItemOption>
+      ))}
+    </MenuOptionGroup>
+  )
+
   return (
-    <AdminLayout title="News">
-      <Box>News</Box>
+    <AdminLayout
+      title="Collections"
```
7alip, line 103: News

libs/utils/src/services/topics/get.ts
```diff
@@ -0,0 +1,56 @@
...
+
+export const topicQueryFn = async () => {
+  const [recommendedTopics, topics] = await Promise.all([
+    getRecommendedTopics(),
+    getTopics(),
+  ])
+  if (!topics) return
+
+  return topics.map(topic => {
+    recommendedTopics.map(recommended => {
```
7alip, line 30: If it doesn't return anything, using `forEach` might make more sense, I think.

libs/utils/src/services/topics/get.ts
```diff
@@ -0,0 +1,56 @@
...
+      }
+    })
+    return topic
+  })
+}
+
+export const useTopic = () => {
+  return useQuery({
+    queryKey: ['topics'],
+    queryFn: topicQueryFn,
```
7alip, line 54: Noticed `Encountered two children with the same key` warnings in the console.

Can you double check if we have duplicated topics in the backend or something wrong in the frontend part?

7alip (commented):
Thanks for great job!

I had to fix typing issues because build process was failing.

`Encountered two children with the same key` issue needs to be fixed.

MustafaTRKYLMZ:
I looked to the topic page. I have commend about view. It might be better when we see news in **drawer** instance of external tab.

7alip:
> I looked to the topic page. I have commend about view. It might be better when we see news in **drawer** instance of external tab.

I noticed that as well. It's probably related to TopicCard issue. Drawer may not be good option as some websites doesn't allow displaying the content in iframe. We can open the page in a new window popup.

libs/utils/src/services/topics/get.ts
rakin-tanis: 👍

apps/admin/pages/content-maker/news.tsx
rakin-tanis: 👍

libs/utils/src/services/topics/get.ts
```diff
@@ -0,0 +1,56 @@
...
+          time: recommended.time,
+          locale: recommended.locale,
+          publisher: recommended.publisher,
+          isRecommended: true,
+        })
+      }
+    })
+    return topic
+  })
+}
```
7alip, line 49: Is the logic something like this?

```ts
return topics.map(topic => {
  const isRecommended = recommendedTopics.some(r => r.url === topic.url)
  
  return { ...topic, isRecommended }
})
```

rakin-tanis: Those two pieces of code don't make the same effect. In the code above, I also add the recommended topics that had been recommended before and are no longer among the new scraping topics. I am not sure that we really need them, but it sounded to me like it is better that way so I added them. if not I can update the code.

7alip: I see it now, thanks. However, I'm not certain if it is important to include outdated topics in this list. We've thought about creating a separate page that would display recommendedTopics only to admins.

apps/admin/pages/content-maker/news.tsx
```diff
@@ -1,12 +1,155 @@
...
+  const [locale, setLocale] = useState<StrapiLocale>(defaultLocale)
+
+  const search = useCallback(
+    (topics: TopicBase[]) => {
+      const results = []
+      topics?.forEach(topicBase => {
+        for (const key in topicBase) {
+          if (
+            typeof topicBase[key] === 'string' &&
+            topicBase[key]?.toLowerCase().includes(searchTerm.toLowerCase())
```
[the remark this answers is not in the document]
7alip: I mentioned regex just because of case insensitivity.

Maybe something like this. But maybe there is no remarkable difference as you said.
```ts
const searchRegex = new Regex(searchTerm, 'gi')
Object.values(topicBase).join(' ').match(searchRegex)
```

7alip (approved):
Looks great, thanks

PR #221, opened 2022-10-16
Onboarding selman

Selman-S:


libs/ui/src/examples/Fetch/E6_RequestUseQuery.tsx
```diff
@@ -2,27 +2,51 @@
...
 
   const fetchData = async () => {
     // TODO: Add request to fetch data
+
+    // eslint-disable-next-line prettier/prettier
+    const data = await Request.collection<StrapiModel[]>({
+      url: url,
+      filters: {
+        [filterProperty]: filterValue,
+      },
```
7alip, line 45: It's easier to test with `containsi`

```suggestion
      filters: {
        [filterProperty]: { $containsi: filterValue },
      },
```

libs/ui/src/examples/TanstackQuery/E3_MultiFetchQuery.tsx
```diff
@@ -31,8 +31,12 @@ const useAuthorsQuery = () => {
 }
 
 const getAuthorBlogs = (authorId?: number) => {
-  return request({
+  return Request.collection<StrapiModel[]>({
     url: 'api/blogs',
+    locale: 'tr',
+    // filters: {
+    //   id: authorId,
+    // },
```
7alip, line 39:
```ts
filters: {
  author: { id: authorId }
}
```

or

```ts
filters: {
  author: { id: { $eq: authorId } }
}
```

libs/ui/src/examples/TanstackQuery/E3_MultiFetchQuery.tsx
```diff
@@ -4,11 +4,11 @@ import { useState } from 'react'
...
 import { Code, Select, Stack } from '@chakra-ui/react'
 import { useQuery } from '@tanstack/react-query'
-import { User } from '@wsvvrijheid/types'
-import { request } from '@wsvvrijheid/utils'
+import { Blog, StrapiModel, User } from '@wsvvrijheid/types'
+import { Request } from '@wsvvrijheid/utils'
 
 const getAuthors = () => {
-  return request<User[]>({
+  return Request.collection<StrapiModel[]>({
```
7alip, line 11: If we know the response type we should specify it. But if the response is dynamically changes by the url, then you can use `StrapiModel`

```suggestion
  return Request.collection<User[]>({
```

libs/ui/src/examples/TanstackQuery/E3_MultiFetchQuery.tsx
```diff
@@ -31,8 +31,12 @@ const useAuthorsQuery = () => {
 }
 
 const getAuthorBlogs = (authorId?: number) => {
-  return request({
+  return Request.collection<StrapiModel[]>({
```
7alip, line 34: The same as above

```suggestion
  return Request.collection<Blog[]>({
```

libs/ui/src/examples/TanstackQuery/E3_MultiFetchQuery.tsx
```diff
@@ -49,13 +53,18 @@ export const MultiFetchQuery = () => {
...
-  console.log('authorsQuery', authorsQuery)
+  console.log(authorsQuery.data?.data)
 
   return (
     <Stack>
       <Select placeholder="Select author">
         {/* TODO: List all the authors and set the selectedAuthorId with onChange method */}
+        {authorsQuery.data?.data.map(author => (
+          // eslint-disable-next-line react-hooks/rules-of-hooks
+          <option onChange={() => setSelectedAuthorId(author.id)}>
```
7alip, line 64: You can use onChange in `Select` element.

`onChange(e => setSelectedAuthorId(e.target.value))`

```suggestion
          <option key={author.id} value={author.id}>
```

7alip (commented):
Thanks for the submission. I noticed some problems needs to be resolved.

libs/ui/src/examples/Fetch/E6_RequestUseQuery.tsx
Selman-S: thanks

PR #225, opened 2022-10-18
Samenvvv

7alip:


7alip:
@MustafaTRKYLMZ After testing we need to create issues for samenvvv. Some parts are still missing, for example tweet search from the backend. There are no posts to be tested in the backend.

MustafaTRKYLMZ:
I tested before, there is a bug I think. Mainhashtag card has to update. (Date etc), If you like it, we can create issues. We should add contact page, about us page etc

PR #229, opened 2022-10-24
Add special characters

yhtyyar94:


7alip (changes requested):

7alip:
We decided to reduce password strength and remove special characters validation. Fixed in e505fd29b0a2219d682923156dcf95fba85dd5d9

PR #231, opened 2022-10-25
Add Cookie Banner Comp.

zahirekrem09:


7alip (changes requested):
Thank you.

`teal` colors could be replaced with `primary` as each site has it's own primary color.

PR #239, opened 2022-10-30
add pop up window to view button

yhtyyar94:


7alip (approved):
Great, thanks!
I've tested `TopicCardBase` in storybook, view button doesn't seem to be working there. Is it because of missing props in the story or something else.

yhtyyar94:
> Great, thanks! I've tested `TopicCardBase` in storybook, view button doesn't seem to be working there. Is it because of missing props in the story or something else.

Yes you are right, I added onView to TopicCardBase stories now

PR #240, opened 2022-10-31
solve bookmark issue

yhtyyar94:


libs/ui/src/admin/TopicCard/TopicCard.tsx
```diff
@@ -24,11 +24,25 @@ export const TopicCard: FC<TopicCardProps> = ({
...
+  )
 
   const handleBookmark = () => {
     if (isBookmarked) {
-      setBookmarksStorage(bookmarksStorage?.filter(url => url !== topic.url))
+      setBookmarksStorage(
+        bookmarksStorage?.filter(
+          url =>
+            url !==
+            (topic.url[topic.url.length - 1] === '/'
```
7alip, line 41: Did you figure out how it's possible that storage urls may have an extra `/` at the end?

7alip (approved):
Thanks

libs/ui/src/admin/TopicCard/TopicCard.tsx
yhtyyar94: > 

no but i thought of such a solution just in case

7alip: @yhtyyar94 I noticed that useLocalStorage from `react-use` doesn't seem to update state after setLocalStorage. I imported it from a new library `usehooks-ts` and it works. Maybe we don't even need to check for `/` anymore.

Another thing I changed: Since all the topics are being replaced with recent data, at some point we may need to store topic object instead of urls to list them in TopicCard.

You can merge after testing.

PR #244, opened 2022-11-04
241-samenvvv-about-us

MSKose:
Note: Don't know which svg icons' colors should be changed. Rest is done, I think.


7alip (approved):
Thank you, do you want to merge it?

PR #266, opened 2022-11-10
feat post create form added

MustafaTRKYLMZ:


apps/samenvvv/pages/about-us.tsx
```diff
@@ -1,19 +1,14 @@
 import React from 'react'
 
-import { Box, Heading, Image, SimpleGrid, Stack, Text, VStack } from '@chakra-ui/react' //Stack,
```
7alip, line 3: Why do we see these about-us changes which is unrelated to this PR?

libs/ui/src/admin/CreateHashtagPostModal/CreateHashtagPostModal.tsx
```diff
@@ -0,0 +1,267 @@
...
+    reset: resetForm,
+  } = useForm<CreateHashtagPostFormFieldValues>({
+    resolver: yupResolver(schema),
+    mode: 'all',
+  })
+
+  const { mutate, isLoading } = useCreateHashtagPost(locale, queryKey)
+  const toast = useToast()
+  const auth = useAuthSelector()
+  console.log('created', auth.user?.id)
```
7alip, line 76: You can remove logs before pushing your code

package.json
```diff
@@ -27,6 +27,7 @@
     "@emotion/styled": "^11.10.5",
     "@hookform/resolvers": "^2.9.10",
     "@mollie/api-client": "^3.6.0",
+    "@mui/material": "^5.10.13",
```
7alip, line 30: For what reason you added `@mui`?

7alip (changes requested):
Thank you very much; excellent work! I'll look into it again extensively after checking all your works quicky.

apps/samenvvv/pages/about-us.tsx
MustafaTRKYLMZ: Because of the lint error. With this error I coudn't push my commit.

package.json
MustafaTRKYLMZ: I don't remember those changes

PR #267, opened 2022-11-10
backend mention search

yuskara:
http://localhost:1337/api/mentions/search?q=samenvvv

### DIFF
--- a/apps/api/src/api/mention/services/mention.ts
+++ b/apps/api/src/api/mention/services/mention.ts
@@ -1,3 +1,4 @@
-import { factories } from '@strapi/strapi';
+import { factories } from '@strapi/strapi'
 
-export default factories.createCoreService('api::mention.mention');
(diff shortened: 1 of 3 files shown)


apps/api/src/api/mention/services/mention.ts
```diff
@@ -1,3 +1,4 @@
-import { factories } from '@strapi/strapi';
+import { factories } from '@strapi/strapi'
 
-export default factories.createCoreService('api::mention.mention');
+// TODO: Add search service when https://github.com/strapi/strapi/pull/14208 is merged
```
7alip, line 3: I would recommend adding "twitterApi" to the mention service but I discovered a problem with TypeScript. 

We can do it later when https://github.com/strapi/strapi/pull/14208 is merged.

7alip (approved):
Thanks, do you want to merge it?

PR #269, opened 2022-11-13
249-topiccard-share-buttons

MSKose:
I couldn't get the placement="bottom" work for Popover, maybe it's some Storybook issue not sure. Rest should be ok


libs/ui/src/admin/TopicCardBase/TopicCardBase.tsx
```diff
@@ -1,13 +1,14 @@
...
-import { ButtonGroup, Stack, Text, useBreakpointValue } from '@chakra-ui/react'
+import { ButtonGroup, Popover, PopoverArrow, PopoverBody, PopoverContent, PopoverTrigger, Stack, Text, useBreakpointValue } from '@chakra-ui/react'
 import { formatDistanceStrict } from 'date-fns'
 import { AiOutlineEye, AiOutlineLike, AiOutlineShareAlt } from 'react-icons/ai'
 import { BsBookmarkHeart } from 'react-icons/bs'
 
 import { WImage } from '../../components'
 import { ActionButton } from './ActionButton'
 import { TopicCardBaseProps } from './types'
+import {ShareButtons} from '@wsvvrijheid/ui'
```
7alip, line 11: You are in `ui` libs folder. NX warns:
`Projects should use relative imports to import from other files within the same project.`

You could import it from components `import { ShareButtons, WImage } from '../../components'`

libs/ui/src/admin/TopicCardBase/TopicCardBase.tsx
```diff
@@ -81,13 +82,29 @@ export const TopicCardBase: FC<TopicCardBaseProps> = ({
...
+            <Popover placement="bottom">
+              <PopoverTrigger>
+                <ActionButton
+                  onClick={() => onShare()}
+                  icon={<AiOutlineShareAlt />}
+                  title="Share"
+                  isVertical={isVertical}
+                  variant="ghost"
+                />
+              </PopoverTrigger>
```
7alip, line 95: There is a placement of the PopoverContent issue. After some investigation I discovered that `ActionButton` (as a children of `PopoverTrigger`) causes the problem.

My guess is: `ActionButton` doesn't receive `ref` attribute as it's not created with `React.forwardRef`.

There seems to be 2 solutions:

1. Refactor ActionButton with `forwardRef`
2. Easier way, just wrap ActionButton with an element, for example `Box` or `div`. So that, PopoverTrigger's children can receive ref.

```tsx
<PopoverTrigger>
  <Box>
    <ActionButton
      onClick={() => onShare()}
      icon={<AiOutlineShareAlt />}
      title="Share"
      isVertical={isVertical}
      variant="ghost"
    />
  </Box>
</PopoverTrigger>
```

7alip (changes requested):
Thanks for the submission. It's almost done. You just need to fix a layout issue I mentioned and also make sure you don't get error when you run `yarn lint`

libs/ui/src/admin/TopicCardBase/TopicCardBase.tsx
7alip: I would use `placement=top` position. What do you think @MustafaTRKYLMZ?

libs/ui/src/admin/TopicCardBase/TopicCardBase.tsx
MSKose: Asking out of curiosity, is there a reason explicit importing is prefered for these cases?

Selman-S (approved):

libs/ui/src/admin/TopicCardBase/TopicCardBase.tsx
MustafaTRKYLMZ: It doesn't work in my local. That is why, I couldn't test.

MustafaTRKYLMZ:
We should send limited text in to the twitter or other platform account. For example, after the 200 charakter we should add "..." It seems better than only word 
image

7alip:
> We should send limited text in to the twitter or other platform account. For example, after the 200 charakter we should add "..." It seems better than only word image

Normally It's not related to this issue. We should open a new issue for this bug. But if @MSKose wants to take this task as well, it's not a problem.

libs/ui/src/admin/TopicCardBase/TopicCardBase.tsx
```diff
@@ -81,13 +93,31 @@ export const TopicCardBase: FC<TopicCardBaseProps> = ({
...
+                  />
+                </Box>
+              </PopoverTrigger>
+              <PopoverContent w="max-content">
+                <PopoverArrow />
+                <PopoverBody>
+                  <ShareButtons
+                    title={topic.title}
+                    url={topic.url}
+                    quote={topic.description?.substring(0, 196) + '...' || ''}
```
7alip, line 115: I mentioned that this feature is unrelated to this PR because we should fix it in `ShareButtons` so that we don't have the same problem in other places where we use sharing buttons.

To truncate contents for Twitter, we use a custom utility function called `makeSocialContent` in ShareButtons. Perhaps you can modify this utility function to include `three dots (...)` at the end.

This function was written for use in our post maker app and to share something on Twitter using share buttons. You may include `three dots (...)` feature to this utility function in your current branch or undo your changes so that you or someone else can deal with it later in another issue.

7alip (approved):
Thanks.

I'm sorry for messing your PR up. :) I left a note about the new feature, you can merge your task if you want to revert your changes regarding three dots. Otherwise take your time to include it into `makeSocialContent` function.

libs/ui/src/admin/TopicCardBase/TopicCardBase.tsx
7alip: It's the first time for me to use monorepo as well. In normal projects you can also specify `import aliases` in tsconfig or jsconfig files. But when it comes to monorepo, it might be confusing to use aliases for internal imports. In our cases, we consider each project (libs) as an npm package 📦.

We could maybe use different alias for internal imports like `~components`. But you have to adjust `eslint` to  recognize your internal imports then.

libs/ui/src/admin/TopicCardBase/TopicCardBase.tsx
7alip: It looks the way better to me, but keep in mind that we can use `truncateText` function for SEO to generate descriptions from content. Let's keep it as it is now.

7alip:
@MSKose I needed to make `pre-commit.sh` executable again and fix the conflicts. You can merge it.

PR #273, opened 2022-11-15
feat timeline add

Selman-S:


libs/ui/src/admin/TimelineBoard/TimelineBoard.stories.tsx
```diff
@@ -0,0 +1,38 @@
+// import { Box, SimpleGrid } from '@chakra-ui/react'
+import { Story, Meta } from '@storybook/react'
+
+import { Container } from '../../components'
+import { TOPICS_MOCK } from '../../mocks'
+import Timeline from './TimelineBoard'
+
+export default {
+  title: 'Admin/TopicCardBase',
```
7alip, line 9: This story should be changed to TimelineBoard.

If you need mock data you can do the following:

1. Add some real twitter users in Strapi dashboard `localhost:1337/admin`
2. Fetch timelines `http://localhost:1337/api/timelines`
3. Create `libs/mock/src/strapi/timeline.ts` and paste fetched data in. (use `Timeline[]` type)

libs/ui/src/admin/TweetCard/TweetCardBase.tsx
```diff
@@ -1,9 +1,7 @@
 import { FC } from 'react'
```
7alip, line 1: We use `TweetCardBase` in `TwitterWidget` and in `PostMaker` component. If you modify this component, it will break PostMaker widget.

It doesn't seem to be straightforward to use `TweetCardBase` for both `TimeLine` and `TweetWidget`.

You can convert TimelineTweet data to Tweet data type.

```ts
interface TimelineTweet {
  id: string;
  text: string;
  media: {
    type: string;
    media_key: string;
    url?: string;
    preview_image_url?: string;
  };
  created_at: string
}
```

```ts
export interface Tweet {
  id: string
  user: {
    name: string
    username: string
    profile: string
  }
  image?: string
  videos?: {
    bitrate?: number
    content_type: string
    url: string
  }[]
  text: string
  likes: number
  retweets: number
}
```

7alip (changes requested):
Thanks for your efforts.

I didn't notice that we have different Tweet types for timeline.

We should change Timeline type as it's tweets type is not compatible with `Tweet`.
You can create a new type called `TimelineTweet`.

7alip (commented):
Thanks, it looks good. We can implement some missing parts later in another issue (onEdit, onSave)

If you have time, you can check if this official library https://github.com/twitter/twitter-text/tree/master/js can help us about styling tweet content.

libs/ui/src/admin/TimelineBoard/TimelineBoard.tsx
```diff
@@ -0,0 +1,65 @@
...
+      rounded="lg"
+      p={4}
+      gap={4}
+      overflowY="auto"
+      shouldWrapChildren={true}
+    >
+      {timelines?.map(timeline => (
+        // eslint-disable-next-line react/jsx-key
+        <Box
+          w="500px"
```
7alip, line 31: React requires unique key prop for the elements in a list!

```suggestion
        <Box
          key={timeline.id}
          w="500px"
```

libs/ui/src/admin/TimelineBoard/TimelineBoard.tsx
```diff
@@ -0,0 +1,65 @@
...
+                <HStack noOfLines={1}>
+                  <Text fontSize={'sm'} fontWeight={'bolder'}>
+                    @{timeline.username}
+                  </Text>
+                </HStack>
+              </HStack>
+            </Box>
+          </Link>
+          {/* I assigned height 700 randomly. What height do you think I should assign? */}
+          <Box overflowY="auto" h="700px">
```
7alip, line 49: I think 700 is fine

libs/ui/src/admin/TimelineTweet/TimelineTweetBase.tsx
```diff
@@ -0,0 +1,177 @@
...
+                  onClick={() => onSave(tweet)}
+                >
+                  Save (Bookmark)
+                </MenuItem>
+              </MenuList>
+            </Menu>
+          )}
+        </HStack>
+        <Text fontSize={'15px'} lineHeight={'20px'} ml={2}>
+          {tweet.text.split('\n').map(a => {
```
7alip, line 82: The result is perfect but I would prefer to create a separate component for this, like TweetContet as we can use it in TweetCard.

libs/ui/src/admin/TimelineTweet/TimelineTweetBase.tsx
```diff
@@ -0,0 +1,177 @@
...
+            return (
+              <>
+                <p></p>
```
7alip, line 85: To give a line-break, I would recommend to use css `white-space: pre-wrap` and `word-wrap: break-word`. Because Twitter itself is using that css rules.

7alip (changes requested):
Thanks for your great work.

PR #295, opened 2022-12-07
272 mskose send email feedback

MSKose:


apps/api/src/api/feedback/content-types/feedback/lifecycles.ts
```diff
@@ -0,0 +1,20 @@
...
+        },
+        populate: ['art'],
+      },
+    )
+
+    strapi.plugins['email'].services.email.send({
+      to: result.createdBy.email,
+      from: 'info@wsvvrijheid.nl',
+      subject: `Your art '${populatedArt[0].art.title}' application has been ${result.status}`,
+      html: `<p><a href="http://localhost:1337/admin/plugins/content-type-builder/content-types/api::feedback.feedback${result.id}">${result.message}</a></p>`,
```
7alip, line 17: We shouldn't redirect the end users to Strapi panel which is only for super admins. I prefer to use env variables for site urls. _(for example KUNSTHALTE_SITE_URL)_ Then we can set different urls for each local, staging and production environments.

```ts

html: `<div>
  <p>Editor note: ${result.message}</p>
  <p>View in your art <a href="${process.env.KUNSTHALTE_SITE_URL}/profile">profile</a></p>
</div>`
```

7alip (changes requested):
Thanks for your submission.

I noticed some points need to be fixed.

7alip (approved):
Thanks, seems perfect to me.

Keep in mind we should add `KUNSTHALTE_SITE_URL` in our api env both in local and in production.

7alip:
Would you like to merge it?

PR #299, opened 2022-12-09
Activity Module

zahirekrem09:


7alip (approved):
To reduce the number of services that do not require arbitrary body data, such as approve, publish, and delete, I placed them in a common folder. For any model, we can use common services.

Apart from that, I prefer page solutions to modals. It simplifies the code, but @MustafaTRKYLMZ will probably prefer the modal solution:) 

For page solution, we could add a language switcher button to the detail page so that users can easily switch between model translations. _(We have localizations data of a model)._ But we can continue after we discuss about it.

I think, we can merge it now and create another branch later if changes are required.

Would you like to merge it?

Issue #243, opened 2022-11-01
Kunsthalte Issues


7alip:
- [x] Backend: Add email notification for artists (approved, rejected)
- [x] Backend: Change art-editor email notifications

MSKose:
Isn't the first issue resolved here https://github.com/wsvvrijheid/monorepo/pull/295 ? If not, which model is this referring to.

Also, is the second issue related to the lifecycle in apps/api/src/api/art/content-types/art/lifecycles.ts ? And in that case, is what's wanted just to alter the notification content in that lifecycle?

Thanks in advance.

7alip:
Hey @MSKose, that's right, you had already fix the first issue, thanks again.

Yes, in the second, we'd like to send an email to the editors whenever a new piece of art is created. However, we are still unsure whether we should send an email to all of the editors or to a single editor at random. For now, we can send emails to all of the editors.

Since we have not completed the roles for the users, I recommend using an env variable for the list of editor emails.

```ts
// EDITOR_EMAILS=editor1@example.com,editor2@example.com...
const editorEmails = process.env.EDITOR_EMAILS?.split(',')

if (editorEmails?.length > 0) {
  // send email to all at once
}
```

PR #305, opened 2022-12-18
#243 done

MSKose:


apps/api/src/api/art/content-types/art/lifecycles.ts
```diff
@@ -1,10 +1,26 @@
...
+
+    // populating artist to use in email subject
+    const populatedArtist = await strapi.entityService.findOne('api::art.art', result.id, {
+      populate: ['artist'],
+    },)
+    const artist = populatedArtist.artist
+
+    if (editorEmails?.length > 0) {
+      strapi.plugins['email'].services.email.send({
+        to: editorEmails,
```
7alip, line 14: Does `to: editorEmails` work with array format? Maybe it expects a string value separated by comma, I'm not sure.

apps/api/src/api/art/content-types/art/lifecycles.ts
```diff
@@ -1,10 +1,26 @@
...
-      from: 'info@wsvvrijheid.nl',
-      subject: `New Art Created ${result.title}`,
-      html: `<p><a href="https://api.wsvvrijheid.nl/admin/content-manager/collectionType/api::art.art${result.id}">${result.title}</a></p>`,
-    })
+  async afterCreate({ result }) {
+    // do not forget to define your comma seperated EDITOR_EMAILS in your local env
+    const editorEmails = process.env.EDITOR_EMAILS?.split(',');
+
+    // populating artist to use in email subject
+    const populatedArtist = await strapi.entityService.findOne('api::art.art', result.id, {
```
7alip, line 7: `populatedArt` could be better since you retrieve it from arts data.

7alip (commented):
Thanks

apps/api/src/api/art/content-types/art/lifecycles.ts
MSKose: I was unsure too as the [docs](https://docs.strapi.io/developer-docs/latest/plugins/email.html#sending-emails-with-a-controller-or-service) stated that it takes `string` but it is working as an `array` indeed.

7alip: So, if you tested it and it worked, you can merge it.

apps/api/src/api/art/content-types/art/lifecycles.ts
MSKose: Now restricted it to `fields: ["name", "username"]`

PR #338, opened 2023-01-12
OgImage

7alip:
You can see the result in admin homepage and post edit page


MustafaTRKYLMZ (approved):

MustafaTRKYLMZ:
We can put limit for caps text. Because long text is not useful in caps. For example, 200-character limit can be good.

7alip:
> We can put limit for caps text. Because long text is not useful in caps. For example, 200-character limit can be good.

We can show the image result while creating a post, so that the user can see if the text overflows.

PR #358, opened 2023-01-20
Recomended tweet page

MustafaTRKYLMZ:
added recommended page


apps/admin/pages/timelines/recommended.tsx
```diff
@@ -0,0 +1,31 @@
...
+  RecommendedTweetCard,
+} from '@wsvvrijheid/ui'
+
+const RecommendedTweetPage = () => {
+  const { data: recommendedTweet, isLoading } = useSearchModel<Timeline>({
+    url: 'api/recommended-tweets',
+  })
+
+  return (
+    <AdminLayout title="Recomended Tweets" isLoading={isLoading}>
```
7alip, line 16: Recommended

libs/ui/src/data/schemas/recomendedTweet.ts
```diff
@@ -0,0 +1,28 @@
+import { RecommendedTweet } from '@wsvvrijheid/types'
+import * as yup from 'yup'
+
+import { FormFields } from '../../admin'
+
+export const recomendedTweetSchema = yup.object({
```
7alip, line 6: recommendedTweetSchema

libs/ui/src/data/schemas/recomendedTweet.ts
```diff
@@ -0,0 +1,28 @@
...
+  mentions: yup.array().of(
+    yup.object().shape({
+      label: yup.string(),
+      value: yup.string(),
+    }),
+  ),
+  media: yup.mixed(),
+})
+
+export const recomendedTweetFields: FormFields<RecommendedTweet> = [
```
7alip, line 17: recommendedTweetFields

libs/ui/src/admin/RecommendedTweetCard/RecommendedTweetCard.tsx
```diff
@@ -0,0 +1,91 @@
...
+} from '@chakra-ui/react'
+import { useRecommendTweet } from '@wsvvrijheid/services'
+import { useAuthSelector } from '@wsvvrijheid/store'
+import { RecommendedTweetCreateInput } from '@wsvvrijheid/types'
+import { TimelineTweet as TimelineTweetType } from '@wsvvrijheid/types'
+
+import { CreateTweetForm } from '../../components'
+import { TimelineLocalTweet, TimelineTweet } from '../TimelineTweet'
+import { RecommendedSocialButtons } from './RecommendedSocialButtons'
+import { RecommendedTweetCardProps, RecommenderType } from './types'
```
7alip, line 17: Eslint complains about the order. This might happen because of newly updated `eslint-plugin-import` package. You can try to reinstall with `yarn` and fix lint issues with `yarn lint`

libs/ui/src/admin/TimelineTweet/types.ts
```diff
@@ -6,7 +6,7 @@ import { ModelImageProps } from '../ModelForm/ModelImage'
 
 export type TimelineTweetProps = {
   tweet: TimelineTweet
-  user: {
+  user?: {
```
7alip, line 9: This doesn't need to be optional. Because TS complains when you use `{user.name}` in your component without optional chaining `(user?.name)`

libs/ui/src/admin/RecommendedTweetCard/RecommendedTweetCard.tsx
```diff
@@ -0,0 +1,91 @@
...
+        <CreateTweetForm
+          onSubmit={handleSubmit}
+          isOpen={isOpen}
+          onClose={onClose}
+          originalTweet={editTweet}
+          isNews={false}
+        />
+      )}
+
+      <HStack spacing={4} align={'start'} bg={'white'} rounded={'md'} p={4}>
```
7alip, line 76: I don't think we need `HStack` here. I would remove it.

1. There is only one child of `HStack`. HStack is being used for its children to be aligned horizontally.
2. TimelineTweet has already padding. Using HStack causes more padding space here.

libs/ui/src/admin/RecommendedTweetCard/RecommendedTweetCard.tsx
```diff
@@ -0,0 +1,91 @@
...
+    onOpen()
+  }
+  return (
+    <Stack
+      spacing={4}
+      // align={'start'}
+      bg={'white'}
+      rounded={'md'}
+      shadow={'sm'}
+      p={4}
```
7alip, line 62: Timeline has padding already. I wouldn't give padding here. You can give padding to social Buttons instead. Maybe wrapping it with box

```tsx
<Box p={4}>
 <RecommendedSocialButtons>...
</Box>
```

libs/ui/src/admin/RecommendedTweetCard/RecommendedSocialButtons.tsx
```diff
@@ -0,0 +1,146 @@
...
+    setConfirmState({
+      isWarning: true,
+      title: 'Delete Tweet',
+      description: 'Are you sure you want to delete this tweet?',
+      buttonText: 'Delete',
+      onConfirm: async () => {
+        deleteModelMutation.mutate(
+          { id },
+          {
+            onSuccess: handleSuccess,
```
7alip, line 56: It's not necessart to create an extra function just for resetting confirm state.

libs/ui/src/admin/RecommendedTweetCard/RecommendedSocialButtons.tsx
```diff
@@ -0,0 +1,146 @@
...
+          <PopoverBody>
+            <ShareButtons
+              // title={tweet?.text}
+              url={tweet?.originalTweet?.media?.url}
+              quote={quoteTweet}
+            />
+          </PopoverBody>
+        </PopoverContent>
+      </Popover>
+      <Tooltip label="delete tweet" hasArrow bg="primary.400">
```
7alip, line 133: Delete tweet

libs/ui/src/admin/RecommendedTweetCard/RecommendedSocialButtons.tsx
```diff
@@ -0,0 +1,146 @@
...
+              console.log('error delete mutation', errors)
+            },
+          },
+        )
+        setConfirmState(undefined)
+      },
+    })
+  }
+
+  const getMentions = () => {
```
7alip, line 67: You can remove the function as you don't use it anywhere

libs/ui/src/admin/RecommendedTweetCard/RecommendedSocialButtons.tsx
```diff
@@ -0,0 +1,146 @@
...
+import {
+  ShareButtons,
+  SocialItem,
+  WConfirm,
+  WConfirmProps,
+} from '../../components'
+import { ActionButton } from '../TopicCard'
+
+export interface RecommendedSocialButtonsProps {
+  items?: SocialItem[]
```
7alip, line 27: You can remove if it's not going to be used

libs/ui/src/admin/RecommendedTweetCard/RecommendedSocialButtons.tsx
```diff
@@ -0,0 +1,146 @@
...
+    tweet?.mentions?.map(mention => {
+      mentions = mentions + ' @' + mention?.username
+    })
+    return mentions
+  }
+  const mentions = getMentions()
+
+  console.log('mentions', mentions)
+
+  const quoteTweet = tweet?.text + ' @' + tweet?.mentions[0]?.username || ''
```
7alip, line 79: You can try something like this?:

```ts
const mentions = tweet?.mentions?.map(mention => `@${mention}`).join(' ')
const quoteTweet = [tweet?.text, mentions].filter(a => !!a).join('\n\n')
```

7alip:
Thanks for great efforts! Could you do the followings:

1. Merge dev into your branch
2. Run `yarn` to install updated packages
3. After changes make sure `yarn lint` and `yarn build` works. I don't know why `yarn tsc` doesn't catch ts errors that results in the build failing.

libs/ui/src/admin/TimelineTweet/types.ts
MustafaTRKYLMZ: There is a type error because user is not optional (in REcommendedTweetCard <TimelineTweet user ={tweet?.recommender} ...>. 
Error here
> Type 'RecommenderType' is not assignable to type '{ name: string; username: string; profile: string; }'.
  Type 'undefined' is not assignable to type '{ name: string; username: string; profile: string; }'.ts(2322)
types.ts(9, 3): The expected type comes from property 'user' which is declared here on type 'IntrinsicAttributes & { css?: Interpolation<Theme>; } & { tweet: TimelineTweet; user: { name: string; username: string; profile: string; }; onSave?: ((data: TimelineLocalTweet) => void) | undefined; onEdit?: ((data: TimelineLocalTweet) => void) | undefined; } & StackProps'

PR #364, opened 2023-01-22
Feat/delete news

MustafaTRKYLMZ:
added delete news


libs/ui/src/admin/TopicCard/TopicCard.tsx
```diff
@@ -239,6 +279,19 @@ export const TopicCard: FC<TopicCardProps> = ({ topic, onCreatePost }) => {
                 colorScheme={topic.isRecommended ? 'green' : 'gray'}
               />
             )}
+            {user && topic?.isRecommended && id && (
```
7alip, line 282: 1. React will not render `undefined` or `null`. You don't need to use ternary with empty string.

```tsx
// Instead of this
{x ? <div /> : ""}

//You can use
{x && <div />}
```

2. News don't have `id` and delete button should be hidden for normal news as they are stored in big JSON data.

7alip (approved):
I hide delete button for normal news.

PR #366, opened 2023-01-22
Post filter by hashtag

MustafaTRKYLMZ:
draff
setPosts() ...


7alip (approved):
I noticed that we had some filtering issues in generic search function.
I added parent relation search for a model.
I updated queryKey of useSearch to make it uniq for different params. Otherwise, event though params are different we would have wrong cache result if queryKey the same.

7alip:
There might some other issues. I'll merge this to complete other refactoring work. We can handle the issues after refactoring changes.

PR #378, opened 2023-02-01
create post from recommended tweet

MustafaTRKYLMZ:
create post from recommended tweet added

### DIFF
--- a/libs/ui/src/admin/RecommendedTweetCard/RecommendedSocialButtons.tsx
+++ b/libs/ui/src/admin/RecommendedTweetCard/RecommendedSocialButtons.tsx
@@ -10,12 +10,14 @@ import {
   PopoverTrigger,
   Tooltip,
 } from '@chakra-ui/react'
+import { API_URL } from '@wsvvrijheid/config'
 import { useDeleteModel } from '@wsvvrijheid/services'
-import { RecommendedTweet } from '@wsvvrijheid/types'
+import { Post, RecommendedTweet } from '@wsvvrijheid/types'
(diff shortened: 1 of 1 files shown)


libs/ui/src/admin/RecommendedTweetCard/RecommendedSocialButtons.tsx
```diff
@@ -97,6 +118,21 @@ export const RecommendedSocialButtons: FC<RecommendedSocialButtonsProps> = ({
...
+      <Tooltip label="Create post" hasArrow bg="primary.400">
+        <ModelCreateModal
+          title="Create Post"
+          url="api/posts"
+          schema={postSchema}
+          fields={postFields}
+          model={postContent}
+          buttonProps={{
+            variant: 'ghost',
+            colorScheme: 'gray',
```
7alip, line 130: We can just add to buttonProps `iconSpacing: isVertical ? 0 : 2`

libs/ui/src/admin/RecommendedTweetCard/RecommendedSocialButtons.tsx
```diff
@@ -97,6 +118,21 @@ export const RecommendedSocialButtons: FC<RecommendedSocialButtonsProps> = ({
...
+          }}
+        >
+          {isVertical ? '' : 'Create Post'}
```
7alip, line 133: I would do `{!isVertical && 'Create Post'}`

7alip (approved):
Thanks, it looks good to me.

I noticed taht we should probably add default `Locale Select` in Edit and Create forms by default. Let's do it in a separate task.

PR #387, opened 2023-02-09
Samen homepage

bllakcn:


apps/samenvvv/pages/index.tsx
```diff
@@ -16,62 +16,100 @@ import i18nConfig from '../next-i18next.config'
...
+              {t('home.post-maker.title')}
+            </Heading>
+            <Text fontSize="xl" fontWeight="normal" maxWidth="2xl">
+              {t('home.post-maker.content')}
+            </Text>
+
+            <Button
+              as={Navigate}
+              href={link || '/'}
+              size={{ base: 'sm', md: 'lg' }}
```
7alip, line 54: I don't recommend using `sm` size for mobile as it might be difficult to click on it for some people. It could even be `lg` on mobile as well.

apps/samenvvv/pages/index.tsx
```diff
@@ -16,62 +16,100 @@ import i18nConfig from '../next-i18next.config'
...
+              fontWeight="semibold"
+              variant="solid"
+              colorScheme="primary.600"
+              bg="white"
+              color="primary.500"
+              boxShadow="lg"
+              _hover={{ color: 'white', bg: 'blackAlpha.100' }}
+            >
+              {t('home.post-maker.button')}
```
7alip, line 63: Since the button text might be long for small devices, we can override `whiteSpace="nowrap"` for this button.

```tsx
size="lg"
whiteSpace="normal"
h="auto"
py="4"
```

Screenshot 2023-02-10 at 01 05 18

apps/samenvvv/pages/index.tsx
```diff
@@ -16,62 +16,100 @@ import i18nConfig from '../next-i18next.config'
...
                 </Text>
-
                 <Button
                   as={Navigate}
                   href={link || '/'}
-                  size="lg"
-                  variant="outline"
-                  colorScheme="primary"
-                  color="white"
+                  size={{ base: 'sm', md: 'lg' }}
```
7alip, line 89: Same as the previous button

apps/samenvvv/pages/index.tsx
```diff
@@ -16,62 +16,100 @@ import i18nConfig from '../next-i18next.config'
...
+                  fontWeight="semibold"
+                  variant="solid"
+                  colorScheme="primary.600"
```
7alip, line 92: Chakra expects color group name (colorScheme="primary") for colorScheme, instead of color shades. You can remove it as it probably has no affect here.

apps/samenvvv/pages/index.tsx
```diff
@@ -16,62 +16,100 @@ import i18nConfig from '../next-i18next.config'
...
+                  {t('read-more')}
                 </Button>
               </Stack>
-
-              <PostMakerIcon
-                order={{ base: 1, lg: 2 }}
-                boxSize={{ base: 300, lg: 500 }}
+              <WImage
+                width={480}
+                height={270}
```
7alip, line 103:
```suggestion
                ratio={16 / 9}
```

Using fixed width will break responsive design for small devices (~360px). By default we use [`fill`](https://nextjs.org/docs/api-reference/next/image#fill) property from Next.js which doesn't require width or height to be set.

Normally, people recommend using width and height for better performance. And we know dimensions of the images coming from our backend upfront. But this is something we should improve later in `WImage` component.

apps/samenvvv/pages/index.tsx
```diff
@@ -16,62 +16,100 @@ import i18nConfig from '../next-i18next.config'
...
+                borderRadius={'xl'}
+                border={'1px'}
+                borderColor={'white'}
+                objectFit="fill"
```
7alip, line 107: I'm not sure `fill` is correct value here. `objectFit="cover"` is already being used by default in WImage.

apps/samenvvv/pages/index.tsx
```diff
@@ -16,62 +16,100 @@ import i18nConfig from '../next-i18next.config'
...
-const Home: FC<HomeProps> = ({ seo, link }) => {
+const Home: FC<HomeProps> = ({ seo, link, hashtag }) => {
   const { t } = useTranslation()
 
   return (
     <Layout seo={seo} isDark hasScroll>
       <Box pos="relative" bg="white" mt="-100px">
-        <Center
+        <Stack
           p={8}
```
7alip, line 29: I noticed that p={8} gives quite large space for small devices. You can maybe wrap Stack with Container instead of using `p={8}`. Or you can use `p={{ base: 4, lg: 8 }}`

```tsx
import { Container } from "@wsvvrijheid/ui"

<Box bgGradient="linear(to-b, primary.600, primary.300)">
  {/* Because we have mt={'-100'} in the parent box  */}
  <Container pt={100}>  
    <Stack py={16}>
```

apps/samenvvv/pages/index.tsx
```diff
@@ -16,62 +16,100 @@ import i18nConfig from '../next-i18next.config'
...
           bgGradient="linear(to-b, primary.600, primary.300)"
           shadow="primary"
           rounded="sm"
```
7alip, line 32: Shadow and rounded are not necessary I think.

7alip (changes requested):
Great job for the first commit! 👍 

I left some comments to improve responsive design.

7alip (approved):
Thanks for the improvements

I was considering to make the first content full height. If you don't mind I adjusted the design a bit. You can review [my changes](https://github.com/wsvvrijheid/monorepo/pull/387/commits/7a2991c084a286854f9549cbcc8cc9a9e585bf18), if you are ok with that, you can merge it.

PR #391, opened 2023-02-13
TimeFormat is changed.

ertugrul-aydin:


apps/samenvvv/components/HashtagCard/HashtagCard.tsx
```diff
@@ -45,7 +45,10 @@ export const HashtagCard: FC<SliderHeroProps> = ({ item, type }) => {
             </Heading>
           )}
         </Box>
-        <Text>{item.date}</Text>
+        <Box>
+          <Icon as={FaCalendarDay} mr={2} />
+          <FormattedDate date={item.date as string} />
+        </Box>
```
7alip, line 51: Nice catch! I forgot that we have `FormattedDate` component 😄 

FaCalendarDay looks a bit bold to me. Can you try `HiOutlineCalendar` instead.

Using HStack (horizontal Flex with gap) might be better for vertical alignment.
```ts
<HStack>
  <Icon />
  <FormattedDate>
</HStack>
```

7alip (changes requested):
Thank you!

apps/samenvvv/components/HashtagCard/HashtagCard.tsx
ertugrul-aydin: I updated icon and box elements 'HStack' instead.

7alip (approved):
Thank you, it looks great now. Would you like to merge it?

PR #393, opened 2023-02-15
Refactor foundation

MustafaTRKYLMZ:


7alip (approved):
Since we don't have a sharper logo of the partner, I reduced the image size.

PR #397, opened 2023-02-19
Fix Mention list bugs

zahirekrem09:


libs/ui/src/post-maker/Mention/MentionList.tsx
```diff
@@ -35,11 +40,23 @@ export const MentionList = () => {
...
   } = useAppSelector(state => state.post)
 
   const dispatch = useAppDispatch()
+  const { query, locale } = useRouter()
   const { t } = useTranslation()
 
+  useEffect(() => {
+    if (!initialMentions || initialMentions.length === 0) {
+      const payload = {
+        slug: query?.['slug'] as string,
```
7alip, line 53: We need to double check if slug is `undefined` will it ignore the filter and fetch all the mentions? You can add `slug: query?.['slug'] || null` just in case.

7alip (approved):
Looks good, thank you

PR #422, opened 2023-02-26
add media_key and variant to getUserTweets.ts

yhtyyar94:


7alip (approved):
Good catch! Thanks a lot. Let's merge it!

PR #434, opened 2023-02-28
Feat/add pwa support (Samen PWA Test)

bllakcn:


7alip:
@bllakcn It seems to generate `sw.js` file in the public folder but I don't know if it's working properly. I don't see any download app button for example.

Let's park this PR for now until we find a proper solution for it.

PR #435, opened 2023-02-28
create cronjob for account statistics

yhtyyar94:


7alip (changes requested):
Thanks for your time and effort!

The goal is to display each account's weekly progress. This model does not appear to be appropriate for it. You can consider adding `Account Records` maybe.

Model can look like

```ts
type AccountRecord = {
  username: string
  tweets: number // Not sure if we can get number of tweets posted by user
  followers: number
  followings: number
  retweets: number
  replies: number
  date: string
}
```

We can use this records in a chart component.

| username | tweets | likes | retweets | date       |
|----------|--------|-------|----------|------------|
| x        | 12     | 44    | 1        | 13-01-2023 |
| y        | 4      | 5     | 0        | 13-01-2023 |
| x        | 5      | 66    | 4        | 20-01-2023 |


### Cron Job

For cron job, you can get account names from env variable

```ts
ACCOUNTS="samenvvv,samenvvven,samenvvvnl"

const accounts = ACCOUNTS.split(',')
```

### Function

You can create a separate function that receives a single account name and returns stats. We can use this function in cron job, services and controllers in the future. For example we can send account name and get the result from api.

apps/api/src/libs/twitter/getAccountStats.ts
```diff
@@ -0,0 +1,45 @@
+import { twitterApi } from './client'
+
+export const getAccountStats = async (
+  username: string,
+  startDate: number = 7,
+) => {
+  const user = await twitterApi.v2.userByUsername(username, {
```
7alip, line 7: You could wrap the function with `try-catch` to prevent the process from crashing. We might get errors from twitter.

apps/api/src/libs/twitter/getAccountStats.ts
```diff
@@ -0,0 +1,45 @@
+import { twitterApi } from './client'
+
+export const getAccountStats = async (
+  username: string,
+  startDate: number = 7,
```
7alip, line 5: `days`, `numberOfDays` or similar name would be more descriptive since the value is not a date type

apps/api/src/libs/twitter/getAccountStats.ts
```diff
@@ -0,0 +1,45 @@
...
+
+  const followerCount = user.data.public_metrics.followers_count
+  const followingCount = user.data.public_metrics.following_count
+  const tweetCount = user.data.public_metrics.tweet_count
+  let likeCount = 0
+  let retweetCount = 0
+  let replies = 0
+
+  const today = new Date()
+  const pastWeek = new Date(today.getTime() - startDate * 24 * 60 * 60 * 1000)
```
7alip, line 20:
```suggestion
const startDate = new Date(today.getTime() - days * 24 * 60 * 60 * 1000)
```

7alip (changes requested):
You should just update cron interval, the rest is not so important but just about naming

7alip (approved):
Thank you

PR #447, opened 2023-03-05
Hashtag anounscements

MustafaTRKYLMZ:


apps/post-maker/.expo/README.md
```diff
@@ -0,0 +1,8 @@
+> Why do I have a folder named ".expo" in my project?
```
7alip, line 1: Expo files seem to be included from other branch and needs to be removed in this PR

apps/samenvvv/components/HashtagAnnouncement/HashtagAnnouncement.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+export const HashtagAnnouncement = ({
+  title,
+  description,
+  date,
+  defaultCaps,
+  hashtag,
+  content,
+  join,
+  link,
```
7alip, line 24: Wouldn't it be simpler just to pass `hashtag` itself as prop. 
You could still generate the link inside the component.

apps/samenvvv/components/HashtagAnnouncement/HashtagAnnouncement.tsx
```diff
@@ -0,0 +1,89 @@
...
+}: HashtagAnnouncementProps) => {
+  const { formattedDate } = useLocaleTimeFormat(date?.value, 'dd MMMM yyyy')
+  const { formattedDate: formatedHour } = useLocaleTimeFormat(
+    date?.value,
+    'HH:mm',
+  )
+  const TurkeyHour = date?.value
+    ? format(addHours(new Date(date?.value as string), 2), 'HH:mm')
+    : undefined
```
7alip, line 33: 1. variable names should be camelCase instead of PascalCase
2.  It's not always 2 hours difference. I would recommend using [date-fns-tz](https://github.com/marnusw/date-fns-tz) for different time zones

apps/samenvvv/pages/announcements.tsx
```diff
@@ -0,0 +1,144 @@
...
+  })
+
+  const latestHashtag = hashtagsQuery?.data?.data?.[0] || null
+  const link = getItemLink(latestHashtag, latestHashtag?.locale, 'hashtag')
+
+  const hashtag = hashtagsQuery?.data?.data?.[0]
+  const hashtagAnounscement = useAnnoucementData(
+    hashtag,
+    locale as StrapiLocale,
+  )
```
7alip, line 46: It looks like a mapper function rather than hook. Therefore the name is confusing. Would be something like `mapHashtagToAnnouncement`

libs/ui/src/hooks/useAnnouncementData.tsx
```diff
@@ -0,0 +1,54 @@
+import { Hashtag, StrapiLocale } from '@wsvvrijheid/types'
+
+export const useAnnoucementData = (hashtag: Hashtag, locale: StrapiLocale) => {
+  const AnnouncementData = {
+    tr: {
+      title: 'ETKİNLİK DUYURUSU',
+      description: { name: 'Konu:', value: hashtag?.description },
+      date: { name: 'Tarih: ', value: hashtag?.date },
```
7alip, line 8: I don't believe we should make it so complicated. I believe hashtag data could be used in the Announcement component without the need for a mapper.

Similar to the following:

```tsx
<Text>
  <strong>{t('topic')}</strong> {hashtag.title}
</Text>
<Text>
  <strong>{t('description')}</strong> {hashtag.description}
</Text>
.
.
.
```

7alip (changes requested):
Thanks for hard work. I think we can simplify it, don't hasitate to ask if you have any question.

apps/samenvvv/pages/announcements.tsx
MustafaTRKYLMZ: Where should we put this function? utils, hooks or?

apps/samenvvv/components/HashtagAnnouncement/HashtagAnnouncement.tsx
```diff
@@ -0,0 +1,42 @@
...
+        <Text>{hashtag?.content}</Text>
+        <Link href={link}>
+          <Text fontWeight={'bold'} color={'primary'} m={4}>
+            {t('join-link')}
+          </Text>
+        </Link>
+        <ShareButtons
+          title={capsParams.title}
+          url={`${SITE_URL}${linkCaps}`}
+          quote={`${capsParams.text}\n\n${hashtag.content}\n\n`}
```
7alip, line 37: I don't think we should share the content.

> We shouldn't render `HashtagAnnouncement` if hashtag is not found. Build seems to fail because of `hashtag.content` where `hashtag` can be undefined.

```suggestion
          quote={`${capsParams.text}\n\n`}
```

apps/samenvvv/pages/announcements.tsx
```diff
@@ -0,0 +1,191 @@
...
+    tr: '@samenvvvTr',
+  }
+
+  if (!hashtag) {
+    return {
+      props: {
+        ...ssrTranslations,
+        seo: { title: title[locale] },
+        hashtag: null,
+      },
```
7alip, line 134: It's my mistake. We shouldn't render the page if hashtag is not found. Otherwise we might show some announcement content without any actual hashtag. This case doesn't make any sense. 

```suggestion
      notFound: true,
```

apps/samenvvv/pages/announcements.tsx
```diff
@@ -0,0 +1,191 @@
...
+}) => {
+  const { t } = useTranslation()
+
+  return (
+    <Layout seo={seo} isDark>
+      <Head>
+        <meta
+          property="twitter:image:src"
+          content={seo.openGraph.images[0].url}
+        />
```
7alip, line 54: Is it because twitter doesn't show the image with `og:image`? I remember had to add it in post page but it seems weird to me. I expect that twitter should show the image with `og:image` just like it shows in our legacy samen app.

apps/samenvvv/pages/announcements.tsx
```diff
@@ -0,0 +1,191 @@
...
+import Head from 'next/head'
+import { serverSideTranslations } from 'next-i18next/serverSideTranslations'
+import { MDXRemoteSerializeResult } from 'next-mdx-remote'
+import { serialize } from 'next-mdx-remote/serialize'
+import { NextSeoProps } from 'next-seo'
+import { useTranslation } from 'react-i18next'
+
+import i18nConfig from '..//next-i18next.config'
+import { Layout } from '../components'
+import { HashtagAnnouncement } from '../components/HashtagAnnouncement' //TODO fix import style
```
7alip, line 29:
```suggestion
import { Layout, HashtagAnnouncement } from '../components'
```

apps/samenvvv/pages/announcements.tsx
```diff
@@ -0,0 +1,191 @@
...
+  mapHashtagToOgParams,
+} from '@wsvvrijheid/utils'
+import { isPast } from 'date-fns'
+import { GetServerSideProps } from 'next'
+import Head from 'next/head'
+import { serverSideTranslations } from 'next-i18next/serverSideTranslations'
+import { MDXRemoteSerializeResult } from 'next-mdx-remote'
+import { serialize } from 'next-mdx-remote/serialize'
+import { NextSeoProps } from 'next-seo'
+import { useTranslation } from 'react-i18next'
```
7alip, line 25:
```suggestion
import { useTranslation } from 'next-i18next'
```

apps/samenvvv/pages/announcements.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+import i18nConfig from '..//next-i18next.config'
```
7alip, line 27:
```suggestion
import i18nConfig from '../next-i18next.config'
```

7alip (changes requested):
Thanks. I noticed some parts that prevents samen from build.

Could you check `yarn build:samenvvv` works before pushing your changes?

PR #449, opened 2023-03-05
Style/post maker responsive

bllakcn:
- Option 1 leaves out the **TrendListTab** in the mobile devices
- Option 2 places **it** under the **PostContainer** in the grid


7alip:
@MustafaTRKYLMZ Could you check these two commits? Which one is preferred?

@bllakcn A screenshot would be easier to compare.

bllakcn:
**OPTION 1**
option1
**OPTION 2**
option2

Issue #432, opened 2023-02-27
The Back button should take you to the parent page rather than the previous one.


7alip:
In model details/edit pages, we use `router.back()` for the Go Back button. For example, changing the model's language on this page takes the user to the target model's details. From `/posts/1` to `/posts/4.` The Back button returns the user to the previous page, `/posts/1`. It would be ideal if it went to the parent posts page `/posts`.

PR #450, opened 2023-03-05
Fix : Admin layout go back bugs

zahirekrem09:
#432

### DIFF
--- a/libs/ui/src/admin/Layout/AdminLayout.tsx
+++ b/libs/ui/src/admin/Layout/AdminLayout.tsx
@@ -31,7 +31,6 @@ export type AdminLayoutProps = {
   isLoading?: boolean
   hasBackButton?: boolean
   seo: NextSeoProps
-  slug?: string
 }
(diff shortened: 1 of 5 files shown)


libs/ui/src/admin/Layout/AdminLayout.tsx
```diff
@@ -113,7 +115,7 @@ export const AdminLayout: FC<AdminLayoutProps> = ({
                         icon={<FaArrowLeft />}
                         colorScheme={'blackAlpha'}
                         rounded="full"
-                        onClick={() => router.back()}
+                        onClick={() => router.push(`/${slug}`)}
```
7alip, line 118: Could it be done without the use of any additional props?
 
```ts
const slug = router.asPath.split("/")[1]
```

7alip (approved):
Thanks

7alip (approved):
Thank you! Would you like to merge?

PR #456, opened 2023-03-11
Bug/art club template filter bug

bllakcn:


7alip (approved):
Thanks, you can merge it.

PR #461, opened 2023-03-13
Admin refactor

MustafaTRKYLMZ:
title deleted, form updated


7alip (approved):
Thanks. I removed the caps line if no title is provided. Actually, it seems that the line will not be used in post caps because the title has already been removed. However, we may use the line in announcement caps.

Aside from that, we may need the title again in the next version of the post. There will be one main post and its alternate posts. The main post can have a title, and we can connect alternates by looking at the post title. Otherwise we will be looking at the id of the post while connecting the alternate posts.

PR #465, opened 2023-03-14
Create custom like and view routes for art and blog

yhtyyar94:


apps/api/src/api/art/controllers/custom.ts
```diff
@@ -12,4 +12,66 @@ export default {
...
     return { data: result }
   },
+  async like(ctx: Context) {
+    if (ctx.state.user) {
+      const art = await strapi.service('api::art.art').findOne(ctx.params.id, {
+        populate: {
+          likers: true,
+        },
+      })
+      const isLiked = art.likers.filter(user => user.id == ctx.state.user.id)
```
7alip, line 22: JS built-in `some` function is more performant than filter. Because it will stop iterating when it finds a matching result.

```ts
const isLiked = art.likers.some(user => user.id == ctx.state.user.id)
```
___

I'm not sure if strapi will always give you all the likers data in this query. There is probably limit to 100 items in a single query. You can try increasing the limit just in case.

```ts
const art = await strapi.service('api::art.art').findOne(ctx.params.id, {
  limit: 1000,
  populate: {
    likers: true,
  },
})
```

Or maybe the following approach is better?

```ts
const likedArts = await strapi.service('api::art.art').find({
  likers: { id: { $eq: ctx.state.user.id } }
})

const isLiked = (likedArts?.data?.length || 0) > 0
```

apps/api/src/api/art/controllers/custom.ts
```diff
@@ -12,4 +12,66 @@ export default {
...
+          .service('api::art.art')
+          .update(ctx.params.id, {
+            data: {
+              likers: likers,
+              likes: art.likes + 1,
+            },
+          })
+
+        return updatedArt
+      } else {
```
7alip, line 37: If you return from an if block, it's not necessary to use else. It improves readability.

apps/api/src/api/art/controllers/custom.ts
```diff
@@ -12,4 +12,66 @@ export default {
...
+      const art = await strapi.service('api::art.art').findOne(ctx.params.id, {
+        populate: {
+          likers: true,
+        },
+      })
+      const isLiked = art.likers.filter(user => user.id == ctx.state.user.id)
+
+      if (isLiked.length === 0) {
+        const likers = [...art.likers]
+        likers.push(ctx.state.user)
```
7alip, line 26: `likers` contains ids of the users. Is `ctx.state.user` is an id or should you push `ctx.state.user.id` here?

7alip (changes requested):
Thanks!

I left some comments. Could you check also `ctx.state.user` when user ids are being used in the query.

apps/api/src/api/art/controllers/custom.ts
yhtyyar94: actually `likers` contains user data in an object not only ids but when we add a new user just id will be enough as u mentioned

PR #466, opened 2023-03-16
 platform muzic and kunsthalte

MustafaTRKYLMZ:
platform muzic added and art station header menu fixed

### DIFF
--- a/libs/config/src/menu/routes.ts
+++ b/libs/config/src/menu/routes.ts
@@ -21,5 +21,5 @@ export const {
   foundation,
   samenvvv,
   terms,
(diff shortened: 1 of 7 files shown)


libs/config/src/menu/routes.ts
```diff
@@ -21,4 +21,5 @@ export const {
   foundation,
   samenvvv,
   terms,
+  muzic,
```
7alip, line 24: Not so important but would be nice if the type is fixed. `music`

7alip (approved):
Thanks

PR #471, opened 2023-03-19
tweet image fixed

MustafaTRKYLMZ:
recommended tweet image fixed

### DIFF
--- a/libs/ui/src/admin/RecommendedTweetCard/RecommendedSocialButtons.tsx
+++ b/libs/ui/src/admin/RecommendedTweetCard/RecommendedSocialButtons.tsx
@@ -60,6 +60,10 @@ export const RecommendedSocialButtons: FC<RecommendedSocialButtonsProps> = ({
     .join(' ')
   const quoteTweet = [tweet?.text, mentions].filter(a => !!a).join('\n\n')
 
+  const tweetImageUrl = tweet?.image
+    ? ''
+    : `https://twitter.com/${tweet.originalTweet?.user?.username}/status/${tweet?.originalTweet?.id}/video/1`
+
   let imageUrl: string | unknown
 
   if (tweet?.image?.url) {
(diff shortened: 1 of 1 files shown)


libs/ui/src/admin/RecommendedTweetCard/RecommendedSocialButtons.tsx
```diff
@@ -60,6 +60,10 @@ export const RecommendedSocialButtons: FC<RecommendedSocialButtonsProps> = ({
     .join(' ')
   const quoteTweet = [tweet?.text, mentions].filter(a => !!a).join('\n\n')
 
+  const tweetImageUrl = tweet?.image
+    ? ''
+    : `https://twitter.com/${tweet.originalTweet?.user?.username}/status/${tweet?.originalTweet?.id}/video/1`
```
7alip, line 65: If users create new tweets by uploading images, we will have images from Strapi without having an original tweet.

If there is no original tweet, the url will be `https://twitter.com/undefined/status/undefined/video/1`. I think we should't generate url if there is no original tweet.

```suggestion
  const tweetImageUrl = tweet?.image && tweet.originalTweet?.user && `https://twitter.com/${tweet.originalTweet?.user?.username}/status/${tweet?.originalTweet?.id}/video/1`
```

7alip (commented):
Thanks

libs/ui/src/admin/RecommendedTweetCard/RecommendedSocialButtons.tsx
7alip: Or maybe its better to check only if original tweet has an image or video?

const tweetImageUrl = tweet?.originalTweet?.image || tweet?.originalTweet?.video && `https://....`

7alip (approved):
Thanks

PR #472, opened 2023-03-21
Twitter accounts

ertugrul-aydin:
Created component. MockData will be added soon.


libs/ui/src/components/AccountStats/AccountStats.tsx
```diff
@@ -0,0 +1,76 @@
...
+        backgroundColor: 'rgba(53, 162, 235, 0.5)',
+      },
+      {
+        label: 'Hesap 3',
+        data: [798, 899, 999],
+        borderColor: '#b856ac',
+        backgroundColor: '#b856ac',
+      },
+    ],
+  }
```
7alip, line 69: This data should be passed through props to make the component dynamic.

We need to create a mapper function that transforms `accountStats` data to `chartData`.

I don't know if the following code works but you can have an idea from it.

```ts
type ChartData = ChartProps<'line', number[]>['data']

const mapAccountStatsToChart = (stats: AccountStats[]): CharData => {
  const labels = stats.map(s => s.date)
  const colors = ["..."]
  const bgs = ["..."]
 
  const dataSets = stats.map((s, index) => ({
    label: s.username,
    data: stats
      .filter(({username}) => userName === s.username))
      .map(({[key]}) => key),
    borderColor: colors[index],
    backgroundColor: bgs[index],
  }))

  return { labels, datasets }

}
```

When you use `AccountStats` component, you can pass the mapped data

```tsx
<SimpleGrid columns={{ base: 1, md:2, lg: 3}}>
  {statsData?.map(stats => <AccountStats data={mapAccounStatsToChart(stats)} />}
</SimpleGrid>
```

7alip (commented):
Thanks

libs/ui/src/components/AccountStats/AccountStats.tsx
7alip: I can the following result from ChatGPT. You can play with it to get the correct result

```tsx
type AccountStats = {
  id: number;
  username: string;
  followers: number;
  tweets: number;
  retweets: number;
  likes: number;
  followings: number;
  replies: number;
  date: string;
  createdAt: string;
  updatedAt: string;
  publishedAt: string;
};

type ChartJSData = {
  labels: string[];
  datasets: Array<{
    label: string;
    data: number[];
    borderColor: string;
    backgroundColor: string;
  }>;
};

export function getChartData(stats: AccountStats[], field: keyof AccountStats): ChartJSData {
  const dataByField = stats.reduce((acc: Record<string, number[]>, curr: AccountStats) => {
    const value = curr[field];
    if (!acc[value]) {
      acc[value] = [];
    }
    acc[value].push(curr[field]);
    return acc;
  }, {});

  const labels = Object.keys(dataByField);
  const datasets = [
    {
      label: field.toString(),
      data: Object.values(dataByField).map((arr) => arr.reduce((a, b) => a + b, 0)),
      borderColor: 'rgba(75,192,192,1)',
      backgroundColor: 'rgba(75,192,192,0.4)',
    },
  ];

  return { labels, datasets };
}
```

apps/admin/pages/index.tsx
```diff
@@ -1,7 +1,8 @@
 import { FC } from 'react'
 
 import { SimpleGrid } from '@chakra-ui/react'
-import { AdminLayout, Caps } from '@wsvvrijheid/ui'
+import { AccountStats } from '@wsvvrijheid/ui'
+import { AdminLayout } from '@wsvvrijheid/ui'
```
7alip, line 5:
```suggestion
import { AccountStats, AdminLayout } from '@wsvvrijheid/ui'
```

apps/api/config/plugins.ts
```diff
@@ -1,6 +1,6 @@
 export default ({ env }) => ({
   'import-export-entries': {
-    enabled: true,
+    enabled: false,
```
7alip, line 3: Why is this change?

libs/ui/src/components/AccountStats/AccountStats.tsx
```diff
@@ -0,0 +1,76 @@
...
+
+  return (
+    <Box p="5" minH={250}>
+      <Line options={options} data={data} />{' '}
```
7alip, line 73:
```suggestion
      <Line options={options} data={data} />
```

7alip (commented):
@ertugrul-aydin I merged dev into your branch to trigger the deployment on Vercel and to fix the conflicts in yarn.lock. Don't forget to pull the changes before adding stuff.

7alip (approved):
Thanks

PR #476, opened 2023-03-25
Art translate

MustafaTRKYLMZ:


apps/admin/pages/arts.tsx
```diff
@@ -55,10 +55,18 @@ const ArtsPage: FC<PageProps> = ({ seo }) => {
...
-  const mappedArts = arts?.map(art => ({
-    ...art,
-    translates: art.localizations?.map(l => l.locale),
-  }))
+  const mappedArts = arts?.map(art => {
+    const translates = []
+
+    if (art.title_en) translates.push('en')
+    if (art.title_tr) translates.push('tr')
+    if (art.title_nl) translates.push('nl')
```
7alip, line 63: It doesn't make sense to check translated fields based on title. Because all title and description fields will be filled on create. To check translation status we could add boolean fields like `isApproved_en`

apps/admin/pages/translates/arts/[id].tsx
```diff
@@ -17,21 +17,21 @@ import i18nConfig from '../../../next-i18next.config'
...
         url="api/arts"
         pathname="arts"
         translatedFields={['title', 'description', 'content']}
         fields={translateModelFields}
         schema={translateModelSchema}
-      />
+      /> */}
     </AdminLayout>
   )
 }
```
Comment (7alip) [inline apps/admin/pages/translates/arts/[id].tsx:37]: We should consider different approach for art translations. Because it doesn't have localizations. For art, translate actually means to update fields.

PR #489, opened 2023-03-29
Feat/share link is localhost in production

MustafaTRKYLMZ:


libs/ui/src/post-maker/PostContainer/PostContainerButtons/PostContainerButtons.tsx
```diff
@@ -33,6 +32,10 @@ export const PostContainerButtons = () => {
   const { isExceeded } = useAppSelector(state => state.post)
 
   const postUrlAbsolute = useItemLink(post as Post, 'post', true)
+  const postUrl = `https://twitter.com/intent/tweet?url=https://samenvvv.nl/${locale}/hashtags/${post?.hashtag?.slug}/${post?.id}&text=${postContent}\n\n\n`
```
7alip, line 35: It doesn't seem a safe link. I recommend using `URLSearchParams`.

```ts
const baseUrl = "https://example.com"
const params = { a: "Hello worl", b: "Good!" }
const query = new URLSearchParams(params)
const result = query.toString()

const url = `${baseUrl}?${result.toString()}
// https://example.com?a=Hello+worl&b=Good%21
```

libs/ui/src/post-maker/PostContainer/PostContainerButtons/PostContainerButtons.tsx
```diff
@@ -33,6 +32,10 @@ export const PostContainerButtons = () => {
...
+
+  console.log('previous url', postUrlAbsolute)
+  console.log('current hashtag', post?.hashtag?.slug)
```
7alip, line 38: You can remove logs

7alip (changes requested):

7alip (approved):
Thanks

PR #496, opened 2023-04-03
Fix disabled chevron buttons

### LINKED ISSUE #478: pagination
When there isn't other art, we can't see next number

image

bllakcn:
fix #478


7alip (approved):
Thanks, nice catch! You can merge it

> I couldn't replicate the issue in #478 _When there is no art available, we shouldn't see the pagination_. Maybe it's not the case anymore

PR #509, opened 2023-04-06
Course types

7alip:
- Updates course and course application model
- Extends course model with curriculum and faqs components
- Adds related types


7alip:
@bllakcn Can you add `instructor` short text field for `Course` model. Remember adding it to CourseBase type as well

bllakcn (approved):

PR #511, opened 2023-04-06
Feat/academy card

bllakcn:
Add AcademyCard


libs/ui/src/academy/AcademyCard/AcademyCard.tsx
```diff
@@ -0,0 +1,49 @@
...
+  children,
+  href,
+  image,
+}) => {
+  return (
+    <Link href={href}>
+      <Card
+        role="group"
+        w={{ base: 48, lg: 64 }}
+        h={{ base: 48, lg: 64 }}
```
7alip, line 23: It's not fully responsive. In smaller devices it should be able to fill the container. Therefore, the Card itself should have 100% width. If you want it always to be square, then you can wrap it with `<AspectRatio ratio={1}>`

By this way, you can easily create a responsive grid by using multiple Cards.

You can also read this approach: https://chakra-ui.com/docs/components/link-overlay/usage#usage-with-routing-library

libs/ui/src/academy/AcademyCard/AcademyCard.tsx
```diff
@@ -0,0 +1,49 @@
...
+          src={image}
+          _groupHover={{ transform: 'scale(1.05 )' }}
+          transition={'all'}
+          transitionDuration={'0.2s'}
+        />
+        <Text
+          _groupHover={{ bottom: 8 }}
+          transition={'all'}
+          transitionDuration={'0.2s'}
+          w={{ base: 48, lg: 64 }}
```
7alip, line 36: You don't need to limit the width for responsive design

libs/ui/src/academy/AcademyCard/AcademyCard.tsx
```diff
@@ -0,0 +1,49 @@
...
+          px="3"
```
7alip, line 37: If horizontal space is 3, it's better to have the same space vertically. (4 might be preffarable)

libs/ui/src/academy/AcademyCard/AcademyCard.tsx
```diff
@@ -0,0 +1,49 @@
...
+        overflow="hidden"
+      >
+        <WImage
+          src={image}
+          _groupHover={{ transform: 'scale(1.05 )' }}
+          transition={'all'}
+          transitionDuration={'0.2s'}
+        />
+        <Text
+          _groupHover={{ bottom: 8 }}
```
7alip, line 33: If you give vertical space with padding py=5 or pb=5, than you could move the element with bottom: 3.

Additionally, this topic may be help to explain why moving elements with translate is better than absolute positioning: https://www.paulirish.com/2012/why-moving-elements-with-translate-is-better-than-posabs-topleft/

apps/foundation/pages/platforms/[slug].tsx
```diff
@@ -23,7 +27,11 @@ const PlatformDetailPage: FC<PlatformTemplateProps> = ({
 
   return (
     <Layout seo={seo}>
-      <PlatformTemplate seo={seo} source={source} image={image} link={link} />
+      {seo.title === 'Wees Academy' ? (
```
Comment (7alip) [inline apps/foundation/pages/platforms/[slug].tsx:30]: seo.title can be translatable and not good for string comparison.

```ts
const router = useRouter()

const slug = router.query.slug // Because page is [slug].tsx
{slug === 'academy' ? ...}
```

apps/foundation/pages/platforms/[slug].tsx
```diff
@@ -23,7 +27,11 @@ const PlatformDetailPage: FC<PlatformTemplateProps> = ({
...
+        <AcademyTemplate seo={seo} image={image} />
+      ) : (
+        <PlatformTemplate seo={seo} source={source} image={image} link={link} />
```
Comment (7alip) [inline apps/foundation/pages/platforms/[slug].tsx:33]: I'm considering to generate this pages on Strapi side. 

You can revert the Template and page related changes and just keep AcademyCard in this PR.

Would also be great to have AcademyCard.stories.tsx.

7alip (changes requested):
Thank you. We are working on dynamic pages which will be managed from Strapi CMS. That's why I would remove the page and template changes.

libs/ui/src/academy/AcademyCard/AcademyCard.tsx
```diff
@@ -0,0 +1,70 @@
...
+              <LinkOverlay as={Link} href={href}>
+                {children}
+              </LinkOverlay>
+              {description && (
+                <Text
+                  display="none"
+                  _groupHover={{ display: 'block' }}
+                  transition="all"
+                  fontWeight="thin"
+                  transitionDuration="0.8s"
```
7alip, line 59: Description can be longer than expected. It's better to limit the number of lines.
```suggestion
                  transitionDuration="0.8s"
                  noOfLines={3}
```

libs/ui/src/academy/AcademyCard/AcademyCard.stories.tsx
```diff
@@ -0,0 +1,35 @@
...
+  image: 'https://placehold.co/256x',
+  href: '/',
+}
+export const withDescription = Template.bind({})
+withDescription.args = {
+  children: 'Courses',
+  image: 'https://placehold.co/256x',
+  href: '/',
+  description:
+    'Lorem ipsum dolor sit amet consectetur, adipisicing elit. Explicabo nulla expedita sunt magnam sequi, at veniam soluta aliquam itaque inventore id. Tempore itaque fuga cupiditate? Ea soluta iste vel sint!,Lorem ipsum dolor sit amet consectetur, adipisicing elit. Explicabo nulla expedita sunt magnam sequi, at veniam soluta aliquam itaque inventore id. Tempore itaque fuga cupiditate? Ea soluta iste vel sint!,',
```
7alip, line 34: This description example seems quite long. You can show the cards in Simple grid to see how the description will look like.

```tsx
const GridTemplate: StoryFn<typeof AcademyCard> = args => {
  return <SimpleGrid columns={{base: 1, md: 2, lg: 4}} gap={4}>
   <AcademyCard {...props} />
    <AcademyCard {...props}
   </SimpleGrid>
}

export const Grid = GridTemplate.bind()
Gird.args = {  ... }
```

7alip (approved):
Thank you so much, looks great!

PR #526, opened 2023-04-09
Dashboard Course Detail

7alip:


7alip:
@MustafaTRKYLMZ I noticed that Course image has multiple option in Strapi. This causes error in ModelEditForm because ModelImage part doesn't support multiple image, it expects the model.image as an object but strapi gives us array course.image.

I updated image.field as single image in this task.

PR #572, opened 2023-05-11
401 user stats

MSKose:


apps/api/src/libs/getStats.ts
```diff
@@ -0,0 +1,167 @@
+export async function getStats(userId: number) {
+  const activityCreatorCount = await strapi.db
+    .query('api::activity.activity')
+    .count({
+      where: {
+        creator: userId,
```
7alip, line 6: Since we run the cron-job on a weekly basis, it's probably best to filter the statistics periodically.

We follow the same approach in `apps/api/src/libs/twitter/getAccountStats.ts`.

```ts
export async function getStats(userId: number, date = new Date(), totalDays = 7) {
    const start_time = formatIso(addDays(date, -totalDays))
    const end_time = formatIso(date)
    
    ...
      where: {
        creator: userId,
        createdAt: { // or date:
          $between: [start_time, end_time] // Not sure if it's a correct syntax
        }
}

**Review by 7alip — COMMENTED** [review]
Thanks, great effort!

I left a comment about filtering the stats. It would be great if you can implement it in a correct way. Otherwise we can improve it later.

File: apps/api/src/libs/getStats.ts (hunk 1 above)
Comment (MSKose) [reply apps/api/src/libs/getStats.ts:6]: Oh, I forgot about that. Thanks for the heads up. I'll resolve this asap.

**Comment by MSKose** [comment]
1. The custom route for individual user's stat request is now set to `path: '/user-statistics/'`. This works fine as the Bearer token is getting accessed from the context in our controller. Getting full list of the users remains to be `path: '/user-statistics/get-stats'`

2. I've changed the $not: {$in: ['public', 'authenticated']} into $notIn: ['public', 'authenticated'] as per the [docs](https://docs.strapi.io/dev-docs/api/entity-service/filter#notin).

File: apps/api/config/env/production/cron-tasks/userStats.ts [hunk 2]
```diff
@@ -0,0 +1,32 @@
...
+import { getStats } from '../../../../src/libs'
+
+export default async ({ strapi }) => {
+  const users = await strapi.entityService.findMany(
+    'plugin::users-permissions.user',
+    {
+      filters: {
+        role: {
+          type: {
+            $notIn: ['public', 'authenticated'],
```
7alip, line 11: Perfect!

apps/api/src/api/user-statistic/routes/custom.ts
```diff
@@ -0,0 +1,14 @@
+export default {
+  routes: [
+    {
+      method: 'GET',
+      path: '/user-statistics/',
```
7alip, line 5: The `find` controller uses this default route. I wouldn't override it because we can use this route in the future with some custom filters.

`find => /user-statistics`
`findOne = > /user-statistics/:id`
`update => /user-statistics/:id`
`delete => /user-statistics/:id`
`create => /user-statistics`

libs/types/src/strapi.ts
```diff
@@ -248,6 +248,8 @@ export type StrapiCollectionUrl =
   | 'timelines'
   | 'tweet-users'
   | 'tweets'
+  | 'user-stats'
+  | 'user-stats/me'
```
7alip, line 252: These routes should be `user-statistics` and the custom one `user-statistics/get-stats`

7alip (commented):
Thanks for the improvements

apps/api/src/api/user-statistic/routes/custom.ts
MSKose: How does this sound?;

For individual requests: `/user-statistics/get-user-stats`
For full list: `/user-statistics/get-stats`

7alip (approved):
Thanks

PR #582, opened 2023-05-17
added static software page

ertugrul-aydin:


apps/foundation/pages/platforms/academy/software.tsx
```diff
@@ -0,0 +1,69 @@
...
+    </Layout>
+  )
+}
+
+export default Platforms
+
+export const getStaticProps = async (context: GetStaticPropsContext) => {
+  const { locale } = context
+
+  const content = {
```
7alip, line 46: I noticed you use `<br />` tag in the content. It's probably best to use Markdown component to render this content. But Markdown component needs serialized markdown source. We use next-mdx-remote.

You can search other pages where we use this approach:

```ts
const source = await serialize(contenxt[locale])
```

apps/foundation/pages/platforms/academy/software.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+type SoftwareProps = InferGetStaticPropsType<typeof getStaticProps>
+
+const Platforms: FC<SoftwareProps> = ({ title, content }) => {
+  const { locale } = useRouter()
+
+  const contentBody = content[locale as StrapiLocale]
+
+  return (
+    <Layout seo={{ title }}>
```
7alip, line 23: If use use Hero component on the page, Layout should take `isDark` prop that changes header menu colors.

```suggestion
    <Layout seo={{ title }} isDark>
```

7alip (changes requested):
Thanks

7alip (changes requested):
Thanks

7alip (approved):
Thanks

PR #595, opened 2023-05-23
Recaptcha

7alip:


libs/ui/src/components/ArtWithDetails/ArtWithDetails.tsx
```diff
@@ -19,43 +17,85 @@ import {
...
+    content,
+    email,
+  }: CommentFormFieldValues) => {
+    if (!art?.id) return
+
+    try {
+      const recaptchaToken = await executeRecaptcha('comment').catch(error => {
+        console.error(error)
+
+        return undefined
```
7alip, line 61: If this part fails, it is best not to perform comment mutation. Because /api/comment always fails without a valid recaptchaToken.

```suggestion
        throw new Error('Recaptcha failed)
```

bllakcn:
Recaptcha seems to be working fine. Yet the comment that is sent does not have an `art` relation attached to it. Thus we don't render any comment. Was this bug previously known?

7alip:
> Recaptcha seems to be working fine. Yet the comment that is sent does not have an `art` relation attached to it in Strapi. Thus we don't render any comment. Though, we have an art id in the payload of the comment requests. I believe this was the root of the issue.

I didn't notice this problem actually. It explains why we don't see them then. We should create a separate task for it. Thanks

7alip:
Since I've been migrating the project to Turborepo, I'm merging this. Please feel free to open a new PR from the same branch.

PR #599, opened 2023-05-25
UserStatistics component is created

yhtyyar94:
a few days ago i upgraded the node version(20) and when i tried to run yarn api it gave me an error, also gave an error when i tried to commit my changes. if there will not occur any conflict it would be better to change the node version in package.json(solved problem by using nvm)


7alip (changes requested):
Thanks. 

1. Can you make it so that multiple users' stats are displayed? I should, for instance, be able to view the statistics of 10 post creators/approvers. Now I can only see 1 user's stats.
2. Component name could be UserStats

libs/ui/src/components/UserStats/UserStatistics.tsx
```diff
@@ -0,0 +1,90 @@
...
+            <XAxis dataKey="date" />
+            <YAxis />
+            <Tooltip />
+            <Legend iconType="square" />
+            <Line
+              type="monotone"
+              dataKey={`stats.${type}.${key}`}
+              stroke={`#${Math.floor(Math.random() * 0xffffff)
+                .toString(16)
+                .padEnd(6, '0')}`}
```
7alip, line 80: We need to make sure if the color is readable. I prefer using theme colors. You can check `getChartData` that we used theme colors.

7alip (approved):
Thanks. We can merge it as the functionality and design works fine. We may do some refactoring in the future. We also need total statistics for all periods displayed in a bar chart, but let's consider adding that in a different task later.

PR #621, opened 2023-05-31
Upload multiple caps

MustafaTRKYLMZ:


7alip (approved):
I created another issue for FilePicker. You can merge it, we will probably improve it later.

PR #644, opened 2023-06-09
fixed blogcard text color

ertugrul-aydin:


packages/ui/src/components/BlogCard/BlogCard.tsx
```diff
@@ -127,7 +127,9 @@ export const BlogCard: FC<BlogCardProps> = ({ post, isFeatured }) => {
           <Heading as="h3" size="md">
             {post.title}
           </Heading>
-          <Text noOfLines={2}>{post.description}</Text>
+          <Text noOfLines={2} color="black">
```
7alip, line 130: Did you figure out what's causing the text's color issue? I'm curious why we need to set the color here. It would be ideal if the underlying problem could be fixed.

If we can't figure out why the color isn't working, I'd rather use the body text color, which is `gray.800`.

7alip: @ertugrul-aydin  I think I understand the problem. BlogCard is wrapped with `Navigate` link component and <a> tags have always [gradient color](https://chakra-ui.com/docs/styled-system/gradient#text-gradient-api).

The solution you already provided is good to me. Just changing it to `gray.800` would bi nice

PR #654, opened 2023-06-14
Update join team

MustafaTRKYLMZ:
age, city and surname added to form


7alip (changes requested):
Thank you.
We need to adjust Strapi Volunteer model. There is no `age` and `city` fields there

7alip (approved):

PR #692, opened 2023-07-02
volunteer and users table added

MustafaTRKYLMZ:


packages/ui/src/data/schemas/user.ts
```diff
@@ -0,0 +1,40 @@
...
+  {
+    name: 'updatedAt',
+    type: 'date',
+    label: 'Updated Date',
+  },
+  {
+    name: 'role',
+    label: 'Role',
+    type: 'select',
+    url: 'api/roles',
```
7alip, line 38: I don't know how to fetch roles via API in Strapi. This url is probably not correct

packages/ui/src/admin/AdminNav/useAdminNav.tsx
```diff
@@ -188,6 +189,18 @@ export const useAdminNav = (user: SessionUser): AdminNavItemProps[] => {
       icon: <GiHumanPyramid />,
       visible: getRoutePermission(roles, '/courses'),
     },
+    {
+      label: tAdmin('volunteers'),
```
7alip, line 193: Missing translations

packages/config/src/data/roles.json
```diff
@@ -0,0 +1,13 @@
...
+  { "id": 4, "name": "ArtEditor" },
+  { "id": 5, "name": "Translator" },
+  { "id": 6, "name": "Admin" },
+  { "id": 7, "name": "ArtEditor Translator" },
+  { "id": 8, "name": "ContentManager" },
+  { "id": "9", "name": "AccountManager" },
+  { "id": 10, "name": "AcademyEditor" },
+  { "id": 11, "name": "Jury" },
+  { "id": 13, "name": "ContentManager Translator" }
+]
```
7alip, line 13: I put hardcoded roles data in config. Because I don't know how to fetch them

packages/ui/src/data/schemas/volunteer.ts
```diff
@@ -0,0 +1,52 @@
...
+    name: 'updatedAt',
+    type: 'date',
+    label: 'Updated Date',
+  },
+  { name: 'age', label: 'Age' },
+  {
+    name: 'jobs',
+    type: 'select',
+    isMulti: true,
+    url: 'api/jobs',
```
7alip, line 49: I'm not able to see Job label in WSelect after updating a volunteer. We should investigate how it's working for categories in some other models

packages/ui/src/data/schemas/user.ts
7alip: Updating role via API doesn't seem to be working. We need another investigation about that.

7alip (commented):
Thanks

packages/ui/src/data/schemas/user.ts
7alip: https://github.com/strapi/strapi/issues/11735#issuecomment-985722733

Strapi doesn't allow updating role by API because of security concerns. We should modify the default user update controller or add a new route/controller for updating user/role.

packages/config/src/data/roles.json
MustafaTRKYLMZ: I think we have create new issue for this

packages/ui/src/admin/AdminNav/useAdminNav.tsx
MustafaTRKYLMZ: > Missing translations

Where should I put translation "volunteers"?

packages/ui/src/data/schemas/user.ts
mikiU2022: Yes, but strapi still keep updateRole on the API ,let us wrongly think we can do this. But no instruction in official document. .... Still don't know how to update the role.

PR #699, opened 2023-07-09
Replace charths with recharts

ertugrul-aydin:


packages/types/src/account-stats.ts
```diff
@@ -2,6 +2,7 @@ import { StrapiBase } from './strapi'
 
 export type AccountStatsBase = {
   username: string
+  dataKey: string
```
7alip, line 5: `AccountStatsBase` is type of the Strapi model. I don't think it has dataKey column and it seems you don't you it anywhere.

7alip (commented):
Thanks. You can merge it after fixing the small changes

7alip (approved):
Thank you

7alip:
@ertugrul-aydin You can merge it

PR #716, opened 2023-07-19
update ModelEditModal

MustafaTRKYLMZ:
- [ ] Change filepicker 
- [x] Show video
- [x] show caps
- [x] show image


7alip (approved):
I know there are a lot of changes for a PR that makes it difficult to review. But if you don't see any important bugs, you can merge it.

MustafaTRKYLMZ:
How can we upload video and caps to other language. We should use caps and video in each language.

PR #718, opened 2023-07-19
music name updated

MustafaTRKYLMZ:
updated music name

### DIFF
--- a/apps/foundation/components/HomePlatform/HomePlatform.tsx
+++ b/apps/foundation/components/HomePlatform/HomePlatform.tsx
@@ -36,14 +36,15 @@ export const HomePlatform: FC<HomePlatformProps> = ({ platforms }) => {
   return (
     <Box>
       {platforms.map((platform, index) => {
-        const color = colors[platform.slug as keyof typeof colors]
+        const color =
+          colors[platform.slug as keyof typeof colors] || colors.academy
 
         return (
           <Center
             key={index}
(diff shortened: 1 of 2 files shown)


apps/foundation/components/HomePlatform/HomePlatform.tsx
```diff
@@ -24,7 +24,7 @@ const colors = {
   samenvvv: { bg: 'samen.100', colorScheme: 'samen', text: 'initial' },
   kunsthalte: { bg: 'green.100', colorScheme: 'green', text: 'initial' },
   academy: { bg: 'blue.100', colorScheme: 'blue', text: 'initial' },
-  scm: { bg: 'red.100', colorScheme: 'red', text: 'initial' },
+  weesmusic: { bg: 'red.100', colorScheme: 'red', text: 'initial' },
```
7alip, line 27: It should be`music` or you need to change `music` to `weesmusic` in `packages/config/src/data/routes.json`

7alip (changes requested):

apps/foundation/components/HomePlatform/HomePlatform.tsx
7alip: It's my mistake. I noticed that in staging api weesmusic slug was still scm, that's why `colors[platform.slug].colorScheme` was failing.

PR #719, opened 2023-07-19
Refactor ModelImage

7alip:


MustafaTRKYLMZ (changes requested):
Filepicker doesn't working in my computer

MustafaTRKYLMZ:
/translates?slug=hashtags&id=43 In translates page normal link. 

image When we click this button link coming like this /hashtags/44 and "not found" response. . I think there is a bug here.

MustafaTRKYLMZ:
image
I saw there problem here. 
1- Preview video is not good, I think. 
2- Caps z-index looks not good. 
3- When I update caps, Image couldn't look good.

PR #725, opened 2023-07-23
onboarding post and form

GOZTURK7:


packages/ui/src/examples/E9_CreateCategoryWithValidation.tsx
```diff
@@ -25,8 +28,8 @@ export const CreateCategoryWithValidation = () => {
     watch,
     setValue,
     formState: { errors },
-  } = useForm<CategoryCreateInput>({
-    // resolver: yupResolver(schema),
+  } = useForm<any>({
```
7alip, line 31: yup has `InferType`, which you can use to determine the schema's type. We created all mutation types for models before, but thanks to InferType, we don't need to define any type for the body if we use schema.

```tsx
const schema = yup.object().shape({
  name_en: yup.string().required(),
  name_nl: yup.string().required(),
  name_tr: yup.string().required(),
  slug: yup.string().required()
})

type CreateCategoryInput = yup.InferType<typeof schema>
```

7alip (approved):
Thanks, nice job!

PR #746, opened 2023-07-30
Fix: Art category filtering

ertugrul-aydin:


7alip (approved):
Thanks, you can merge it.

(Preview urls doesn't work well because staging db has expired. I'll fix it soon but you can merge it)

PR #751, opened 2023-08-02
Show author in blog-edit-form/ last branch after deleting the other two

GOZTURK7:


http.rest
```diff
@@ -0,0 +1 @@
+GET http://localhost:1337/api/authors
```
7alip, line 1: Could you remove `http.rest` file

7alip (approved):
You can merge it after removing the file.

PR #764, opened 2023-08-06
add google pay and apple pay to donation page

yhtyyar94:


apps/dashboard/public/.well-known/apple-developer-merchantid-domain-association
```diff
@@ -0,0 +1 @@
...33303131303630333535303430613063306134313730373036633635323034393665363332653331306233303039303630333535303430363133303235353533303230383463333034313439353139643534333633303064303630393630383634383031363530333034303230313035303061303831393533303138303630393261383634383836663730643031303930333331306230363039326138363438383666373064303130373031333031633036303932613836343838366637306430313039303533313066313730643331333933303338333133393331333733313332333333303561333032613036303932613836343838366637306430313039333433313164333031623330306430363039363038363438303136353033303430323031303530306131306130363038326138363438636533643034303330323330326630363039326138363438383666373064303130393034333132323034323062303731303365313430613462386231376262613230316130336163643036396234653431366232613263383066383661383338313435633239373566633131333030613036303832613836343863653364303430333032303434363330343430323230343639306264636637626461663833636466343934396534633035313039656463663334373665303564373261313264376335666538633033303033343464663032323032363764353863393365626233353031333836363062353730373938613064643731313734316262353864626436613138363633353038353431656565393035303030303030303030303030227D
```
7alip, line 1: I added certificates mentioned here: https://stripe.com/docs/stripe-js/elements/payment-request-button?client=html#verifying-your-domain-with-apple-pay

apps/dashboard/pages/api/payment/index.ts
```diff
@@ -32,7 +32,7 @@ const handler: NextApiHandler = async (req, res) => {
     }
     // Create checkout session for subscription
     const payment = await stripe.checkout.sessions.create({
-      payment_method_types: [method],
+      payment_method_types: [method === 'ideal' ? 'ideal' : 'card'],
```
7alip, line 35: I think we can remove the iDeal and Visa/Maestro buttons from our donation pages. Alternatively, those icons could remain just for information and not be clickable. Users can choose a payment method on the Stripe checkout page.

What do you think @yhtyyar94  @MustafaTRKYLMZ ?

yhtyyar94: yes it will better and easier :)

apps/dashboard/public/.well-known/apple-developer-merchantid-domain-association
yhtyyar94: i already tried it but i could not test it, i need a device with apple pay

apps/dashboard/pages/api/payment/index.ts
7alip: If @MustafaTRKYLMZ agree with that, you adjust the Icon box to non-clickable elements.

apps/dashboard/public/.well-known/apple-developer-merchantid-domain-association
yhtyyar94: i have tested apple pay and it works

apps/dashboard/pages/api/payment/index.ts
```diff
@@ -32,7 +32,6 @@ const handler: NextApiHandler = async (req, res) => {
     }
     // Create checkout session for subscription
     const payment = await stripe.checkout.sessions.create({
-      payment_method_types: [method],
       line_items: [
```
7alip, line 35: If we don't specify `payment_method_types` it seems to display more methods than we expect. I think `payment_method_types: ["ideal", "card"]` is good enough.

Screenshot 2023-08-06 at 21 54 15

7alip (approved):
Thanks. I just removed Buttons from payment icons since they are not clickable anymore.

You can merge it.

PR #770, opened 2023-08-06
Fix localized slugs and LocaleSwitcher

7alip:


ertugrul-aydin (commented):
Everything looks good. Filtering is working now when changing language on Club Page.

PR #773, opened 2023-08-07
implemented apify for timelines

yhtyyar94:


apps/api/config/cron/index.ts
```diff
@@ -22,7 +22,7 @@ export default {
   timelines: {
     task: timeline,
     options: {
-      rule: '0 */6 * * *',
+      rule: '0 */23 * * *',
```
7alip, line 25: Why 23 hours? Could it be `0 0 * * *` everyday at 00:00?

package.json
```diff
@@ -41,6 +41,7 @@
     "@wsvvrijheid/config": "*",
     "@wsvvrijheid/types": "*",
     "@wsvvrijheid/ui": "*",
+    "apify-client": "^2.7.1",
```
7alip, line 44: If we use `apify-client` only in Strapi app, I would add it to apps/api/package.json

7alip (commented):
Excellent! Thanks

yhtyyar94:
all possible i will make the changes

7alip:
Thanks, you can merge it

PR #779, opened 2023-08-08
draft recommendedtweet fixed

MustafaTRKYLMZ:
draft fixed

### DIFF
--- a/packages/ui/src/admin/ModelForm/ModelCreateForm.tsx
+++ b/packages/ui/src/admin/ModelForm/ModelCreateForm.tsx
@@ -115,7 +115,7 @@ export const ModelCreateForm = <T extends StrapiModel>({
     const bodyData = {
       ...body,
       slug,
-      publishedAt: url !== 'api/recommended-topics' ? null : new Date(),
+      publishedAt: url !== ('api/recommended-topics' ) ? url !=='api/recommended-tweets'?null:new Date() : new Date(),
       locale,
     } as StrapiTranslatableCreateInput
 
--- a/packages/ui/src/admin/ModelForm/ModelCreateForm.tsx
+++ b/packages/ui/src/admin/ModelForm/ModelCreateForm.tsx
@@ -33,6 +33,7 @@ export const ModelCreateForm = <T extends StrapiModel>({
   model,
   onSuccess,
   hideLanguageSwitcher,
+  shouldPublish
 }: ModelCreateFormProps<T>) => {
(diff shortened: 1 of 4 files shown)


packages/ui/src/admin/ModelForm/ModelCreateForm.tsx
```diff
@@ -115,7 +115,7 @@ export const ModelCreateForm = <T extends StrapiModel>({
     const bodyData = {
       ...body,
       slug,
-      publishedAt: url !== 'api/recommended-topics' ? null : new Date(),
+      publishedAt: url !== ('api/recommended-topics' ) ? url !=='api/recommended-tweets'?null:new Date() : new Date(),
```
7alip, line 118: Wouldn't it be better to control publish state via props?. We can pass `shouldPublish` to `ModelCreateModal` and `ModalCreateForm`

```suggestion
      publishedAt: shouldPublish ? new Date() : null,
```

7alip (changes requested):
Thanks.

```json
{
  "editor.formatOnSave": true
}
```

Your code was not formatted. We have this settings in `.vscode/settings.json` but it seems your editor doesn't apply this. Maybe `Format on save` checkbox is not active in your VSCode user settings

packages/ui/src/admin/ModelForm/ModelCreateForm.tsx
```diff
@@ -115,7 +116,7 @@ export const ModelCreateForm = <T extends StrapiModel>({
     const bodyData = {
       ...body,
       slug,
-      publishedAt: url !== ('api/recommended-topics' ) ? url !=='api/recommended-tweets'?null:new Date() : new Date(),
+      publishedAt: shouldPublish?new Date() : null,
```
7alip, line 119: 1. `shouldPublish` is optional, so it does not have to be set to 'false'.

`publishedAt: undefined ? new Date() : null` is the same with
`publishedAt: false ? new Date() : null`

2. For boolean props (as true) the following examples are the same:

`<Component shouldPublish />`
`<Component shouldPublish={true} />`

7alip (commented):
Thank you

packages/ui/src/admin/ModelForm/ModelCreateForm.tsx
7alip: Basically, you can just pass `shouldPublish` when it's necessary. You don't need to indicate when it's `false`.

PR #782, opened 2023-08-09
Aktivities in platforms website

MustafaTRKYLMZ:


apps/api/src/api/activity/content-types/activity/schema.json
```diff
@@ -107,6 +113,27 @@
...
+      },
+      "type": "boolean"
+    },
+    "forLotus": {
+      "pluginOptions": {
+        "i18n": {
+          "localized": true
+        }
+      },
+      "type": "boolean"
```
7alip, line 136: Why we have both `platforms` relation and `forLotus`, `forKunsthalte` booleans.

`forPlatform` boolean approach was an alternative way to relation. We should chose one of them. I prefer using `platforms` relation approach.

packages/ui/src/data/schemas/activity.ts
```diff
@@ -22,6 +24,8 @@ export const activityFields: FormFields<Activity> = [
 
   { name: 'content', isRequired: true, type: 'markdown' },
   { name: 'image', type: 'file', isRequired: true },
+  { name: 'forKunsthalte', type: 'boolean',label: 'Publish to Kunsthalte', },
+  { name: 'forLotus', type: 'boolean',label:'Publish to Lotus vd Media' },
```
7alip, line 28: If we use `platforms` relation, these fields should be replaced with `platforms` multi select.

packages/types/src/activity.ts
```diff
@@ -6,6 +6,8 @@ import { Tag } from './tag'
 
 type ActivityBase = StrapiEntityBase & {
   date: string
+  forKunsthalte:boolean
+  forLotus:boolean
 }
 
 type ActivityRelation = {
```
7alip, line 13: Did you forget to add `platforms` relation as type?

MustafaTRKYLMZ: > Did you forget to add `platforms` relation as type?
I don't understand that. I thought that forKunsthalte and forLotus are enought.

7alip: In the [old issue](https://github.com/wsvvrijheid/monorepo/issues/403) I meantioned `forPlatform` approach but in my opinion `platforms` approach is better as we discussed in the [new issue](https://github.com/wsvvrijheid/monorepo/issues/758).

`platforms` relation is more flexible and in the future, if there are more platforms we can easily update platform relation.

apps/foundation/pages/platforms/index.tsx
```diff
@@ -12,6 +12,8 @@ import { HomePlatform, Layout } from '../../components'
 type PlatformsProps = InferGetStaticPropsType<typeof getStaticProps>
 
 const Platforms: FC<PlatformsProps> = ({ title, platforms }) => {
+  console.log('platforms BBBBBB', platforms)
```
7alip, line 15: Console could be removed

apps/kunsthalte/pages/activities/index.tsx
```diff
@@ -0,0 +1,116 @@
...
+    locale,
+    page,
+    filters: { platforms: { id: { $in: [1] } } },
+  })
+
+  const { data, isLoading } = activitiesQuery
+
+  const pagination = data?.meta?.pagination
+  const activities = data?.data || []
+  console.log('aktivities :::', activities)
```
7alip, line 50: Console could be removed

packages/types/src/activity.ts
```diff
@@ -13,12 +14,14 @@ type ActivityRelation = {
...
   image?: UploadFile
   localizations?: Array<Activity>
+  platforms?: Array<Platform>
 }
 
 type ActivityRelationInput = {
   categories?: number
   tags?: Array<number>
   image: File
+  platforms?: number
```
7alip, line 24:
```suggestion
  platforms?: number[]
```

packages/ui/src/admin/ModelForm/utils.ts
```diff
@@ -152,6 +153,13 @@ export const useDefaultValues = <T extends StrapiModel>(
               value: c.id.toString(),
             })) || []
           break
+        case 'platforms':
+          defaults.platforms =
+            activityModel?.platforms?.map(c => ({
```
7alip, line 158:
```suggestion
            activityModel?.platforms?.map(p => ({
```

7alip (changes requested):
Thanks, great!

7alip:
@MustafaTRKYLMZ You can merge after finishing the small changes

apps/foundation/pages/platforms/index.tsx
7alip, line 15: Is the console still there?

apps/kunsthalte/pages/activities/index.tsx
```diff
@@ -0,0 +1,117 @@
...
+
+const args: RequestCollectionArgs = {
+  url: 'api/activities',
+  sort: ['date:desc'],
+  filters: {
+    approvalStatus: { $eq: 'approved' },
+    platforms: { id: { $in: [1] } },
+  },
+  populate: ['image', 'platforms'],
+  fields: ['title', 'description', 'slug'],
```
7alip, line 30: It's better not to use fields here. We might need date or other data in the future for cards

apps/foundation/pages/platforms/index.tsx
MustafaTRKYLMZ: I deleted. But Maybe I forgot.

apps/kunsthalte/pages/activities/index.tsx
MustafaTRKYLMZ: It was before. If it is not nessesary, I can delete.

apps/foundation/pages/platforms/index.tsx
7alip: It seems you forgot to delete it.

apps/kunsthalte/pages/activities/index.tsx
7alip: Not necessary for now. It might be necessary when we update the Card

PR #789, opened 2023-08-13
donations-list

GOZTURK7:


packages/ui/src/data/tables/donation.tsx
```diff
@@ -0,0 +1,32 @@
...
+    componentProps: { format: 'dd MMMM' },
+    sortable: true,
+  },
+  status: {
+    type: 'badge',
+    componentProps: value => {
+      const colorScheme = {
+        approved: 'green',
+        pending: 'yellow',
+        rejected: 'red',
```
7alip, line 23: `aproved` `pending` `rejected`: We use these approval statuses for our translatable data (such as activities, blogs, hashtags etc). 

The type of donation status is different than ApprovalStatus and it's a result of [Stripe webhooks]([Stripe webhooks](https://github.com/wsvvrijheid/monorepo/blob/c5630fdaf538bb0a12ed0016461dcf885c579113/packages/stripe/src/donation/webhook.ts#L9C14-L9C14)).  If the status is paid, we can change the colour to green; otherwise, the status could be yellow.

```ts
return {
    variant: 'outline',
    colorScheme: (value as string) === 'paid' ? 'green' : 'yellow',
}
```

packages/ui/src/admin/AdminNav/useAdminNav.tsx
```diff
@@ -194,5 +194,10 @@ export const useAdminNav = (): AdminNavItemProps[] => {
       link: '/donation',
       icon: <BsCashCoin />,
     },
+    {
+      label: tAdmin('donationList'),
```
7alip, line 198: Do we have this `donationList` key in `admin.json`? 

I would change `donationList` to `donations`;

Each project has public/locales folder. Whenever we add a new translation, we need to make sure if the key exists in `admin.json` or `common.json` for **every project**.

```
const { t } = useTranslation() // => common.json
const { t: tAdmin } = useTranslation('admin') // => admin.json
```

apps/dashboard/pages/donations.tsx
```diff
@@ -0,0 +1,106 @@
...
+  const status = query.status as ApprovalStatus
+
+  const [sort, setSort] = useState<Sort>()
+
+  const { locale } = useRouter()
+
+  const donationsQuery = useStrapiRequest<Donation>({
+    url: 'api/donates',
+    page: currentPage || 1,
+    pageSize: 10,
```
7alip, line 30: You can increase page size, maybe 50 is fine

apps/dashboard/pages/donations.tsx
```diff
@@ -0,0 +1,106 @@
...
+    // filters: {
+    //   ...(status ? { approvalStatus: { $eq: status } } : {}),
+    //   ...(searchTerm && { [`title_${locale}`]: { $containsi: searchTerm } }),
```
7alip, line 33: You can activate filter but only for email search is enough for now.

```ts
...(searchTerm && { email: { $containsi: searchTerm } })
```

apps/dashboard/pages/donations.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+type PageProps = InferGetStaticPropsType<typeof getStaticProps>
+
+const DonationsPage: FC<PageProps> = ({ seo }) => {
+  const { query } = useRouter()
+  const [currentPage, setCurrentPage] = useState<number>(1)
+  const [searchTerm, setSearchTerm] = useState<string>()
+
+  // Client side query params (?status=pending)
+  const status = query.status as ApprovalStatus
```
7alip, line 21: status filter can be removed

apps/dashboard/pages/donations.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+  const donations = donationsQuery?.data?.data
+  const totalCount = donationsQuery?.data?.meta?.pagination?.pageCount || 0
+
+  const mappedDonations = donations?.map(donation => {
+
+    return {
+      ...donation,
+    }
+  }) as Donation[]
```
7alip, line 56: You return the donation data without modifying it. Mapping has no effect in this situation and could be removed.

apps/dashboard/pages/donations.tsx
```diff
@@ -0,0 +1,106 @@
...
+        onSearch={handleSearch}
+        searchPlaceHolder={'Search arts by title or artist'}
+        sortMenu={[
+          <MenuItem key="asc" icon={<FaArrowUp />}>
+            Name Asc
+          </MenuItem>,
+          <MenuItem key="desc" icon={<FaArrowDown />}>
+            Name Desc
+          </MenuItem>,
+        ]}
```
7alip, line 70: `sortMenu` could also be removed

apps/dashboard/pages/donations.tsx
```diff
@@ -0,0 +1,106 @@
...
+      />
+      <DataTable
+        columns={donationColumns}
+        data={mappedDonations}
+        // onSuccess={artsQuery.refetch}
```
7alip, line 75: comments could be removed

7alip (changes requested):
Thanks

packages/ui/src/data/tables/donation.tsx
```diff
@@ -0,0 +1,26 @@
...
+
+// import { LocaleBadges} from '../../admin'
+import { WTableProps } from '../../components'
+
+export const donationColumns: WTableProps<Donation>['columns'] = {
+  amount: { sortable: true },
+  email: {},
+  // translates: {
+  //   transform: value => <LocaleBadges locales={value as StrapiLocale[]} />,
+  // },
```
7alip, line 11: Comments could be removed

packages/ui/src/data/tables/donation.tsx
```diff
@@ -0,0 +1,26 @@
+import { Donation } from '@wsvvrijheid/types'
+
+// import { LocaleBadges} from '../../admin'
```
7alip, line 3: Comments could be removed

packages/ui/src/data/tables/donation.tsx
```diff
@@ -0,0 +1,26 @@
...
+import { WTableProps } from '../../components'
+
+export const donationColumns: WTableProps<Donation>['columns'] = {
+  amount: { sortable: true },
+  email: {},
```
7alip, line 8: Shall we make `email` sortable as well?

7alip (approved):
You can merge after removing the comments

7alip:
@GOZTURK7 Remember adding keys to json files

packages/ui/src/data/tables/donation.tsx
```diff
@@ -5,7 +5,7 @@ import { WTableProps } from '../../components'
 
 export const donationColumns: WTableProps<Donation>['columns'] = {
   amount: { sortable: true },
-  email: {},
+  email: {sortable: true},
```
7alip, line 8: It's not formatted!

MustafaTRKYLMZ:
Is it possible to change sort type? Last donation must be top of the table. Default sort esc can be better.

apps/dashboard/pages/donations.tsx
```diff
@@ -0,0 +1,88 @@
...
+    url: 'api/donates',
+    page: currentPage || 1,
+    pageSize: 50,
+    filters: {
+      ...(searchTerm && { email: { $containsi: searchTerm } }),
+    },
+    sort,
+  })
+
+  useEffect(() => setCurrentPage(1))
```
7alip, line 37: > If `useEffect` hook doesn't have any dependency array, the effect function will be called on every render of the component. This can lead to performance issues and unexpected behavior. It is recommended to always specify the dependencies array to ensure that the effect is only executed when the dependencies have changed. If there are no dependencies, you can pass an empty array [] to ensure that the effect is only executed once, when the component mounts.

```suggestion
  useEffect(() => setCurrentPage(1), [])
```

apps/dashboard/pages/donations.tsx
```diff
@@ -0,0 +1,88 @@
...
+  donationColumns,
+} from '@wsvvrijheid/ui'
+
+type PageProps = InferGetStaticPropsType<typeof getStaticProps>
+
+const DonationsPage: FC<PageProps> = ({ seo }) => {
+  const [currentPage, setCurrentPage] = useState<number>(1)
+  const [searchTerm, setSearchTerm] = useState<string>()
+
+  const [sort, setSort] = useState(['createdAt:desc'] as any)
```
7alip, line 23: You can merge after this adjustment

```suggestion
  const [sort, setSort] = useState<Sort>(['createdAt:desc'])
```

7alip (approved):
Thanks

PR #795, opened 2023-08-15
Feedback table

MustafaTRKYLMZ:


apps/dashboard/pages/userfeedbacks/[id].tsx
```diff
@@ -0,0 +1,78 @@
+import { FC } from 'react'
```
Comment (7alip) [inline apps/dashboard/pages/userfeedbacks/[id].tsx:1]: Shall we use Modal instead of a new page just like we use in some other pages.

Feedback is a pretty simple data, showing in a modal would be better.

7alip (commented):
Thanks

7alip:
@MustafaTRKYLMZ You can also first try to implement it in `[model].tsx` page.

PR #798, opened 2023-08-16
plus button and feedback button added

MustafaTRKYLMZ:


packages/ui/src/components/UserFeedback/UserFeedback.tsx
```diff
@@ -0,0 +1,27 @@
...
+      as={Button}
+      top={'50%'}
+      position={'fixed'}
+      zIndex={'modal'}
+      right={0}
+      onClick={onOpen}
+      isDisabled={isOpen}
+      colorScheme="primary"
+      aria-label="Plus"
+      transform="rotate(270deg)"
```
7alip, line 21: https://feedbackbutton.io/

I found an example here. They use `writing-mode: vertical-rl;` and `rotate(180deg). Our button is not 0 to right side. We could fix it with changing `right={}` but it seems writing-mode solution is cleaner. We could also use svg or image but I prefer this solution.

We should also remove radius. If you still want to give rounded corners, you can apply it to only the open side (left side of the button)

packages/ui/src/components/UserFeedback/UserFeedback.tsx
```diff
@@ -0,0 +1,27 @@
...
+import { Box, Button, useDisclosure } from '@chakra-ui/react'
+
+import { UserFeedbackForm } from '../UserFeedbackForm'
+
+export const UserFeedback = () => {
+  const { isOpen, onOpen, onClose } = useDisclosure()
+
+  return (
+    <Box
+      as={Button}
```
7alip, line 12: It doesn't have to be Box, could be Button.

packages/ui/src/components/UserFeedback/UserFeedback.tsx
```diff
@@ -0,0 +1,27 @@
...
+      top={'50%'}
+      position={'fixed'}
+      zIndex={'modal'}
+      right={0}
+      onClick={onOpen}
+      isDisabled={isOpen}
+      colorScheme="primary"
+      aria-label="Plus"
```
7alip, line 20: It's not needed or you can give more descriptive value, like Give feedback

7alip (commented):
Thanks

apps/samenvvv/components/PlusButton/PlusButton.tsx
```diff
@@ -0,0 +1,66 @@
...
+        placement="top"
+        onOpen={setIsEditing.on}
+        onClose={setIsEditing.off}
+      >
+        <PopoverTrigger>
+          <Button
+            position={'fixed'}
+            zIndex={'modal'}
+            right={0}
+            bottom={50}
```
7alip, line 38: - [ ] It shouldn't be right={0} I think. right and bottom can have the same value (4, 8, 12 etc)
- [ ] It could be `IconButton` with large size and isRound. `Button` is not circle now.
- [ ] Plus icon could be thicker, FaPlus or any other icon

packages/ui/src/components/UserFeedback/UserFeedback.tsx
```diff
@@ -0,0 +1,28 @@
...
+      h={'fit-content'}
+      position={'fixed'}
+      zIndex={'modal'}
+      right={0}
+      onClick={onOpen}
+      isDisabled={isOpen}
+      colorScheme="primary"
+      aria-label="Give feedback"
+      transform="rotate(180deg)"
+      sx={{ writingMode: 'vertical-rl' }}
```
7alip, line 22: Do you think it should be rounded? No round (rounded={0}) would be nicer in my opinion. Or at least borderBottomRightRadius, borderBottomLeftRadius could be 0.

apps/samenvvv/components/PlusButton/PlusButton.tsx
MustafaTRKYLMZ: Removing right=0 is not possible with this style. Other why, It is not possible to fit right side. Do you have any other idea for that?

7alip: I mean, Plus button doesn't need to be adjacent to the right side of the page. It can be better if the bottom and right spaces are the same (`bottom={x} right={x}`. But for other Feedback button, `right={0}` is perfect

packages/ui/src/components/UserFeedback/UserFeedback.tsx
```diff
@@ -0,0 +1,30 @@
...
+      borderBottomRightRadius={0}
+      borderBottomLeftRadius={0}
```
7alip, line 24: My mistake, probably these are the correct values

```
borderTopLeftRadius: 0;
borderBottomLeftRadius: 0;
```

Screenshot 2023-08-17 at 13 35 57

7alip (approved):
Thanks

PR #807, opened 2023-08-24
platform field added to volunteer edit form

GOZTURK7:


packages/ui/src/data/schemas/volunteer.ts
```diff
@@ -55,4 +56,10 @@ export const volunteerFields: FormFields<Volunteer> = [
     url: 'api/users',
   },
   { name: 'comment', label: 'Comment' },
+  {
+    name: 'platform',
```
7alip, line 60:
```suggestion
    name: 'platforms',
```

7alip (changes requested):
Thanks

packages/types/src/volunteer.ts
```diff
@@ -25,7 +25,7 @@ export type VolunteerBase = {
   occupation: string | null
   phone: string | null
   twitter: string | null
-  platform: string | null
+  platforms: string | null
```
7alip, line 28: It should be `platforms?: number[]` and it's not a base type but a input type. It should be in `VolunteerRelation` type

7alip: We use these input types as body type for post or put mutations.

```ts
const inputVolunterBody = {
  ...
  platforms: [1, 5, 6] // number[] for ids
}
```

7alip: In the same way, you can add `volunteers: Array<Volunteer>` relation type and `volunteers?: Array<number>` relation input type in platforms.ts type.

7alip (approved):
Thanks, you can merge

PR #826, opened 2023-09-10
Refactor strapi url and role permissions

7alip:
- Uses `endpoint` instead of `url` (`api/models` becomes `models`)
- Removes StrapiUrl type
- Allows the adjusting of action permissions (the "approve, delete, edit, publish, and reject buttons") from a single configuration (`packages/config/src/roles`, `packages/ui/src/hooks/usePermission`)


packages/config/src/roles.ts
```diff
@@ -0,0 +1,69 @@
+import { PartialStrapiEndpointMap, RoleType } from '@wsvvrijheid/types'
+
+export type RoleActionType =
+  | 'create'
+  | 'approve'
+  | 'update'
+  | 'delete'
+  | 'publish'
+
+export const actionRolesEndpoints: PartialStrapiEndpointMap<
```
7alip, line 10: @MustafaTRKYLMZ These roles should be extended.

PR #831, opened 2023-09-12
Allow reading hashtag sentences for public

7alip:
I noticed that post maker page gives error while fetching hashtag sentences:

```json
{  "error": "Cannot read properties of undefined (reading 'length')" }
```

Hopefully it fixes the error


yhtyyar94 (approved):
Thanks, good catch. I was getting this error too but somehow it was working.

PR #839, opened 2023-09-16
Playwright

yhtyyar94:


package.json
```diff
@@ -14,7 +14,9 @@
     "clean": "turbo run clean && rm -rf node_modules",
     "storybook": "turbo run storybook --concurrency=20",
     "storybook-build": "turbo run storybook-build",
-    "format": "prettier --write \"**/*.{ts,tsx,md,json}\""
+    "format": "prettier --write \"**/*.{ts,tsx,md,json}\"",
+    "playwright": "kill-port 3000 3001 3002 3003 3004 1337 && playwright test",
```
7alip, line 18: Duplicate

```suggestion
    "playwright": "yarn kill-port && playwright test",
```

7alip (approved):
Thanks it seems to work. We can merge it and try to make it work in Github actions in a separate branch.

package.json
yhtyyar94: The yarn kill-port command alone does not terminate the ports; we also need to specify the port.

7alip: It's probably because you use the same script as the package's script. This would work:

`kill-ports` => your custom script
`kill-port` => package's script

```
    "playwright": "yarn kill-ports && playwright test",
    "kill-ports": "yarn kill-port && playwright test",
```

yhtyyar94: I have read the docs and tried both options unfortunately this package does not have such property. We need to specify the port.

PR #840, opened 2023-09-16
Add Total Amount Feature to Donations Page

ertugrul-aydin:


apps/dashboard/pages/donations.tsx
```diff
@@ -81,7 +86,15 @@ const DonationsPage = () => {
         currentPage={currentPage}
         setCurrentPage={setCurrentPage}
         onSort={setSort}
-      />
+      >
```
7alip, line 89: You don't need to show the children if `totalAmount` doesn't exist

apps/dashboard/pages/donations.tsx
```diff
@@ -81,7 +86,15 @@ const DonationsPage = () => {
...
+        <Spacer />
+        <Flex textAlign={'right'}>
+          <Spacer />
+          <Box paddingY={2} paddingX={5} bg="white" shadow="base">
+            Total Amount: <Text as="b">{totalAmount}</Text>
+          </Box>
+        </Flex>
```
7alip, line 96: Having `Spacer`s along with `textAlight=right` seems overkill to me.

Can something like this work:

```tsx
<Flex justify="end">
  <Box />
</Flex>
```

7alip (commented):
Thanks

packages/ui/src/admin/DataTables/DataTable.tsx
```diff
@@ -23,6 +24,7 @@ export const DataTable = <T extends StrapiModel>({
           <Box>{t('no-data')}</Box>
         )}
       </Box>
+      {tableProps.data.length > 0 && children}
```
7alip, line 27: What if we want to display something as DataTable children even though there is no data?

apps/dashboard/pages/donations.tsx
7alip: I would control displaying totalAmount here instead of `{tableProps.data.length > 0 && children}`.

ertugrul-aydin: Yes, you are right. i will make the change.

packages/ui/src/admin/DataTables/DataTable.tsx
ertugrul-aydin: Yes, you are right. i will make the change.

apps/dashboard/pages/donations.tsx
```diff
@@ -1,6 +1,6 @@
 import { useEffect, useState } from 'react'
 
-import { useUpdateEffect } from '@chakra-ui/react'
+import { useUpdateEffect, Box, Spacer, Text, Flex } from '@chakra-ui/react'
```
7alip, line 3: Lint error, Spacer is defined but never used

```suggestion
import { useUpdateEffect, Box, Text, Flex } from '@chakra-ui/react'
```

7alip (approved):
Thanks

7alip:
> If I see correct, amount show paid and unpaid amount. If it true, It should be changed. We need only paid amount.

@MustafaTRKYLMZ Shouldn't we adjust it with status filter?  I don't know if it makes sense but we can still see what is the total amount of unpaid transactions.

ertugrul-aydin:
> If I see correct, amount show paid and unpaid amount. If it true, It should be changed. We need only paid amount.

I've updated it to display both 'paid' and 'unpaid' amounts separately. What do you think about this ? I have shared screenshot below.

Ekran Resmi 2023-09-18 12 32 46

7alip:
> I've updated it to display both 'paid' and 'unpaid' amounts separately. What do you think about this ? I have shared screenshot below.

It's also fine for me, thanks

MustafaTRKYLMZ:
> > If I see correct, amount show paid and unpaid amount. If it true, It should be changed. We need only paid amount.
> 
> @MustafaTRKYLMZ Shouldn't we adjust it with status filter? I don't know if it makes sense but we can still see what is the total amount of unpaid transactions.

In mijn opinion We don't need unpaid amount. I think paid amount is enought. For example one person can try lots of time.

apps/dashboard/pages/donations.tsx
```diff
@@ -64,6 +64,21 @@ const DonationsPage = () => {
...
+        return acc + (donation.amount || 0)
+      }, 0)
+
+  const totalUnpaid =
+    donations &&
+    donations
+      .filter(donation => donation.status !== 'paid')
+      .reduce((acc, donation) => {
+        return acc + (donation.amount || 0)
+      }, 0)
```
7alip, line 81: Our conclusion is to list only paid transactions. So we don't need this part

apps/dashboard/pages/donations.tsx
```diff
@@ -81,7 +96,17 @@ const DonationsPage = () => {
...
         setCurrentPage={setCurrentPage}
         onSort={setSort}
-      />
+      >
+        {donations && (
+          <Flex justify={'end'}>
+            <Box paddingY={2} paddingX={5} bg="white" shadow="base">
+              Total Paid: <Text as="b">{totalPaid}</Text>
+              <br />
+              Total Unpaid: <Text as="b">{totalUnpaid}</Text>
```
7alip, line 105: Remove unpaid part

apps/dashboard/pages/donations.tsx
```diff
@@ -64,6 +64,21 @@ const DonationsPage = () => {
 
   const donations = donationsQuery?.data?.data as Donation[]
   const totalCount = donationsQuery?.data?.meta?.pagination?.pageCount || 0
+  const totalPaid =
```
7alip, line 67: Update donationsQuery filter, only fetch them based on paid status.

`totalPaid` can be renamed to `totalAmount`

7alip (changes requested):
We decided to display always only paid donations. Can you make the changes?

7alip (commented):
Thank you,

There are some conflicts. Would you prefer to resolve them yourself?

PR #846, opened 2023-09-18
create yml file for github actions

yhtyyar94:


package.json
```diff
@@ -45,6 +45,7 @@
     "@wsvvrijheid/ui": "*",
     "axios": "^1.5.0",
     "chakra-react-select": "^4.7.2",
+    "ci": "^2.3.0",
```
7alip, line 48: Does it required?

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Yarn install
+        run: npm install -g yarn
+      - name: Install dependencies
+        run: yarn install
+      - name: Install dependencies
+        run: yarn ci
+      - name: Install Playwright Browsers
+        run: yarn playwright install --with-deps
+      - name: Run Playwright tests
+        run: yarn playwright test
```
7alip, line 25: I see that the [pipeline](https://github.com/wsvvrijheid/monorepo/actions/runs/6226209561/job/16898264097 ) is failing when it tries to run Strapi server. 

I'm not sure if it's possible to run both api and frontend through the pipeline without deploying it. Frontend needs to connect to the backend while testing.

We can try changing env variables. API_URL will not be localhost but the staging endpoint for testing. We should also avoid running strapi dev. `yarn dev` will run every app, we can add a new script that triggers only frontend apps' dev script.

7alip (commented):
Thanks

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,45 @@
...
+  push:
+    branches: [playwright-github-actions]
+  pull_request:
+    branches: [playwright-github-actions]
+  deployment_status:
+
+jobs:
+  get_branch_name:
+    runs-on: ubuntu-latest
+
```
7alip, line 12: @yhtyyar94 Can we enable Turbo caching?

See bottom of the page: https://turbo.build/repo/docs/ci/github-actions

```
    env:
      TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
      TURBO_TEAM: ${{ vars.TURBO_TEAM }}
```

yhtyyar94: We can try it at another branch after completing the current issue.

packages/playwright/config.ts
```diff
@@ -0,0 +1,25 @@
+const isDevMode = process.env['isDev']
```
7alip, line 1: Can you check this? I think most people use that variable

```suggestion
const isDevMode = process.env['NODE_ENV'] === 'development'
```

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,30 @@
+name: Playwright Tests
+on:
+  push:
+    branches: [dev]
+  pull_request:
+    branches: [dev]
```
7alip, line 6: Whenever we merge some task branches into dev, we will trigger test.
I think we can trigger a new test once we create a pull request from dev to main.

```suggestion
    branches: [main]
```

packages/playwright/config.ts
7alip: @yhtyyar94 or maybe it's better to use `process.env.CI`

packages/playwright/config.ts
```diff
@@ -0,0 +1,25 @@
...
+type platform =
```
7alip, line 2: I prefer using PascalCase for type names

```suggestion
type Project =
// or type Site =
```

packages/playwright/config.ts
yhtyyar94: I will try CI

packages/playwright/config.ts
```diff
@@ -1,4 +1,4 @@
-const isDevMode = process.env['isDev']
+const isDevMode = process.env['CI']
```
7alip, line 1: @yhtyyar94 `process.env['CI']` doesn't mean dev mode. You can consider it as the opposite of dev mode.

Url will be Vercel url for CI, otherwise it will be localhost.

7alip: Since we don't use automated tests in production, CI is like production mode for tests. Otherwise we assume it's in dev mode.

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,30 @@
+name: Playwright Tests
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
```
7alip, line 6: I meant changing only `pull_request` to main. `push` could remain for dev.

```suggestion
  push:
    branches: [dev]
  pull_request:
    branches: [main]
```

7alip (approved):
Thanks for great effort. We can merge it.

PR #849, opened 2023-09-20
onboarding task #3

hakanayata:
1. Worked on the same branch as the last one
2. Assumed at least one field would be required to submit the form

### DIFF
--- a/packages/ui/src/examples/E1_FetchWithAxios.tsx
+++ b/packages/ui/src/examples/E1_FetchWithAxios.tsx
@@ -10,21 +10,16 @@ import { TOKEN } from '@wsvvrijheid/secrets'
 import { Blog } from '@wsvvrijheid/types'
 
 
-// const BLOG_URL = `${API_URL}/api/blogs` // https://api.wsvvrijheid.nl/api/blogs?locale=tr
-const BLOG_URL = `https://api.wsvvrijheid.nl/api/blogs?locale=tr` // https://api.wsvvrijheid.nl/api/blogs?locale=tr
+const BLOG_URL = `${API_URL}/api/blogs?locale=tr`
 
 const headers = { Authorization: `Bearer ${TOKEN}` }
 
 const getBlogs = async () => {
   try {
     const res = await axios.get(`${BLOG_URL}`, { headers })
-    if (res.status === 200) {
-      console.log(res.data)
+    if (res.status !== 200) throw new Error("Failed to fetch")
 
-      return res.data.data
-    } else {
-      console.error("Failed to fetch: ", res.statusText)
-    }
+    return res.data.data
   } catch (error) {
     console.error("Error fetching data: ", error)
   }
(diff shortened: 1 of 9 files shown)


packages/ui/src/examples/E1_FetchWithAxios.tsx
```diff
@@ -1,23 +1,57 @@
...
 import { useRouter } from 'next/router'
 
 import { API_URL } from '@wsvvrijheid/config'
 import { TOKEN } from '@wsvvrijheid/secrets'
+import { Blog } from '@wsvvrijheid/types'
 
-const BLOG_URL = `${API_URL}/api/blogs` // https://api.wsvvrijheid.nl/api/blogs?locale=tr
+
+// const BLOG_URL = `${API_URL}/api/blogs` // https://api.wsvvrijheid.nl/api/blogs?locale=tr
+const BLOG_URL = `https://api.wsvvrijheid.nl/api/blogs?locale=tr` // https://api.wsvvrijheid.nl/api/blogs?locale=tr
```
7alip, line 14: It's my mistake. We should never use production api in our stories, especially for mutations.

```
const BLOG_URL = `https://wsvv-api-staging.onrender.com/api/blogs?locale=tr`
```

You can also change NEXT_PUBLIC_API_URL to https://wsvv-api-staging.onrender.com in your `packages/ui/.env.local`

Alternatively, you can run your local backend api with `yarn --cwd=apps/api dev`, it will be available from `http://localhost:1337`

packages/ui/src/examples/E6_CreateCategory.tsx
```diff
@@ -29,12 +31,28 @@ export const CreateCategory: FC<CreateCategoryWithAxiosProps> = ({
...
     // TODO: Provide category body
+    const categoryBody = {
+      slug,
+      name_en,
+      name_nl,
+      name_tr,
+    }
+
+    // ? I assumed that at least one field would be required?
+    const isFormValid = slug.trim() || '' && name_en.trim() || '' && name_tr.trim() || '' && name_nl.trim() || ''
```
7alip, line 42: Normally each field should be required but I don't remember how the backend validation was. It's ok.

7alip (commented):
Thanks

packages/ui/src/examples/E1_FetchWithAxios.tsx
```diff
@@ -1,23 +1,57 @@
...
+const headers = { Authorization: `Bearer ${TOKEN}` }
+
+const getBlogs = async () => {
+  try {
+    const res = await axios.get(`${BLOG_URL}`, { headers })
+    if (res.status === 200) {
+      console.log(res.data)
+
+      return res.data.data
+    } else {
```
7alip, line 25: Again, `else` is not necessary after an early return if statement.

7alip: To reduce indentation and improve readability you can return for error cases first.

```ts
if (res.status !== 200) throw Error()

// Rest of your code
```

packages/ui/src/examples/E6_CreateCategory.tsx
```diff
@@ -29,12 +33,27 @@ export const CreateCategory: FC<CreateCategoryWithAxiosProps> = ({
     let blogs
 
     // TODO: Provide category body
+    const categoryBody = {
```
7alip, line 36: Almost every model has a mutation input type. If you use these types, TS will assist you with autocomplete and pre-validation. In this case, `CategoryCreateInput` could be used. `CategoryUpdateInput`, for example, is used for update mutations.

7alip (approved):
Perfect, thank you!

7alip (approved):
Thanks, perfect!

PR #853, opened 2023-09-23
Fix updating user role

7alip:
I noticed that `/users` endpoint expects body without `{ data: ... } ` field unlike the other endpoints.

I'm still unable to update `avatar`, we should probably adjust backend controller for that.


7alip:
Implemented in #854

PR #901, opened 2023-10-03
The responsive card completed.

hasancomce:


packages/ui/src/examples/E0_ResponsiveCard.tsx
```diff
@@ -4,28 +4,33 @@ export const ResponsiveCard = () => {
...
+      boxShadow="md"
+      bg="white"
+    >
       <Stack>
-        <Image src={randomPic} alt="random pic" />
+        <Image
+          src={randomPic}
+          alt="randomPicDesc"
+          objectFit="cover"
+          w={{ base: '70%', md: '40%' }}
```
7alip, line 19: On desktop, `Image` will be 40% of the parent `Stack` component. On desktop, I would give the parent a specific pixel width and have the image fill in the parent vertically and horizontally `(h="full")`.

```tsx
<Stack w={{ base: "full", md: "150px" }}  flexShrink={0} />
  <Image w={{ base: "70%", md: "full" }} h={"full"} />
```

Screenshot 2023-10-04 at 04 29 18

___

It is not mentioned, but you could also change the HTML structure, remove Image's parent stack; styling would be easier in this case.

packages/ui/src/examples/E0_ResponsiveCard.tsx
```diff
@@ -4,28 +4,33 @@ export const ResponsiveCard = () => {
...
+          m={{ base: 'auto', md: 'none' }}
```
7alip, line 20: I usually use the Stack (Flex) properties (`justify` and `align`) to control alignments. If there is no flex container, using margin is still a good idea.

packages/ui/src/examples/E0_ResponsiveCard.tsx
```diff
@@ -4,28 +4,33 @@ export const ResponsiveCard = () => {
...
+          aspectRatio={{ base: 1, md: 'none' }}
```
7alip, line 21: 😱 Didn't notice this feature. Perfect! Thank you for allowing me to learn something new 😄

7alip (commented):
Thanks. Only the image size on desktop needs to be improved to match the design, the rest looks perfect.

PR #910, opened 2023-10-08
Onboarding/post and form examples hasancomce

hasancomce:


packages/ui/src/examples/E6_CreateCategory.tsx
```diff
@@ -48,10 +55,21 @@ export const CreateCategory: FC<CreateCategoryWithAxiosProps> = ({
         <Input
           placeholder="Category name (en)"
           value={name_en}
-          onChange={e => setNameEn(e.target.value)}
+          onChange={(e: any) => setNameEn(e.target.value)}
```
7alip, line 58: Why it's `any`? Do you get TS errors without any? When I hover on on it I see: 

`(parameter) e: ChangeEvent<HTMLInputElement>`

packages/ui/src/examples/E7_CreateCategoryWithUseMutation.tsx
```diff
@@ -39,10 +52,20 @@ export const CreateCategoryWithUseMutation = () => {
           <Input
             placeholder="Category name (en)"
             value={name_en}
-            onChange={e => setNameEn(e.target.value)}
+            onChange={(e: any) => setNameEn(e.target.value)}
```
7alip, line 55: The same, it shouldn't require type definition here.

packages/ui/src/examples/E8_CreateCategoryWithHookForm.tsx
```diff
@@ -23,20 +23,23 @@ export const CreateCategoryWithHookForm = () => {
 
   useEffect(() => {
     // TODO: Update slug with slugify on name_en change
-  }, [])
+    setValue('slug', slugify(name_en ?? ''))
+  }, [name_en])
 
   const onSubmit = async (data: CategoryCreateInput) => {
-    const body = { ...data, slug: slugify(data.name_en ?? '') }
+    // const body = { ...data, slug: slugify(data.name_en ?? '') }  // Line26'da setValue kullanildigi icin bu satira gerek kalmadigini dusunuyorum.
```
7alip, line 30: Agree!

packages/ui/src/examples/E6_CreateCategory.tsx
hasancomce: I got an error without 'any'. But now I delete it and the error is gone too. I think vs code was not working properly before. No problem now. thank you.

PR #911, opened 2023-10-09
Fix join form design and add isVolunteer field

7alip:


7alip:
@MustafaTRKYLMZ It should send an email once a profile is created with `isVolunteer` field:

`apps/api/src/api/profile/content-types/profile/lifecycles.ts`

```ts
export default {
  afterCreate({ result }) {
    if (result.isVolunteer) {
      strapi.plugins['email'].services.email.send({
        to: 'info@wsvvrijheid.nl',
        from: 'info@wsvvrijheid.nl',
        subject: `New volunteer ${result.name}`,
        html: `<pre>${JSON.stringify(result, null, 2)}</pre>`,
      })
    }
  },
}
```

PR #913, opened 2023-10-11
Onboarding/e4 and e5 hasancomce

hasancomce:


packages/ui/src/examples/E5_FetchWithCustomQuery.tsx
```diff
@@ -3,12 +3,35 @@
...
+    // sort: ['date:desc'],
+    filters: {
+      title: { $containsi: title },
+    },
+    // pageSize: 1,
+  }
+
+  const queryKey = Object.values(args)
+
+  await queryClient.prefetchQuery(queryKey, () => strapiRequest<Blog>(args))
```
7alip, line 29: We use prefetchQuery for hydrated state in server side of the page. `fetchBlogs` function is just a fetcher function of `useQuery` and it could only have `fetch`, `axios` or preferably  `strapiRequest` to get the data.

The main idea of this task is to create a custom fetch hook instead of using `useQuery` directly in the component.

Using useQuery
```ts
const blogQuery = useQuery({
  queryKey: [],
  queryFunction: () => {}
})
```
Using custom hook
```ts
// Create a custom hook outside of the component
const useCustomQuery = () => {
  return useQuery({
  queryKey: [],
  queryFunction: () => {}
})

// Use the hook in any component
const Component = () => {
  const customQuery = useCustomQuery()
  
  return <pre>{JSON.stringify(customQuery.data, null, 2)}</pre>
}
```

7alip (commented):
Thank you

packages/ui/src/examples/E5_FetchWithCustomQuery.tsx
hasancomce: I completely misunderstood the task. I am editing and resubmitting in the light of your warnings. Thank you.

7alip (approved):
Thanks

PR #919, opened 2023-10-22
Donation FilterMenu is added

ertugrul-aydin:


7alip:
@ertugrul-aydin I had a new thought after viewing your code, thanks. The component is now more generic. You can review and merge it if agree.

7alip:
@ertugrul-aydin If you want, you can try to removing `donations.tsx` page and get the donations table work in `[endpoint].tsx`

7alip:
> @7alip initially 'paid' is selected, but the incoming data contains all data. I added the 'All' option for startup. Is it better this way? Also, I think 'unpaid' does not come from Api.

I think the only important status is `paid`. Other statuses is not so important but maybe someone may want to check some transactions. That's why I wanted to be able to see all the status.

I prefer showing only `paid` transactions initially.

And, yes, `unpaid` does exists in the api:

https://dashboard-git-donations-filters-wsvvrijheid.vercel.app/donations?status=unpaid

ertugrul-aydin:
I have reverted it back to its previous state. At the beginning, 'paid' is preselected. It might be better if 'defaultValue' is left empty.
Ekran Resmi 2023-10-24 23 29 11

7alip:
@ertugrul-aydin Preselected `paid` option is fine. Having `all` option would be nice, `all` means clear the filter and show all, I think.

7alip (approved):
Thanks, can you merge it?

PR #927, opened 2023-10-24
Fix profile

MustafaTRKYLMZ:


packages/ui/src/components/JoinForm/schema.ts
```diff
@@ -31,6 +31,7 @@ export const joinSchema = () => {
     isPublic: yup.boolean(),
     availableHours: yup.number().min(1).max(40).required(),
     heardFrom: yup.array().required().min(1),
+    platforms: yup.array(),
```
7alip, line 34: We don't get `platforms` from form fields. So, should be removed from schema.

packages/ui/src/components/JoinForm/JoinForm.tsx
```diff
@@ -49,8 +49,27 @@ export const JoinForm: FC<JoinFormProps> = ({
...
+              platformsId.push(platform?.id)
+          }
+        })
+      })
+    })
+    const newPlarformsId = platformsId.join().split(',')
+
+    const newData = { ...data, platforms: newPlarformsId }
+
+    onSubmitHandler(newData)
```
7alip, line 70: Is this part working?

I noticed Strapi changed the way of connecting multiple relations with API
https://docs.strapi.io/dev-docs/api/rest/relations

I tried this but it didn't work:

```ts
const onSubmit: SubmitHandler<JoinFormFieldValues> = data => {
    const platformIds: number[] = []
    const jobIds = data.jobs.map(id => parseInt(id))

    platforms.forEach(platform => {
      const selectedJobIds = data.jobs.map(id => parseInt(id))
      const isPlatformSelected = selectedJobIds.some(id => {
        return platform?.jobs?.some(job => job.id === id)
      })

      if (isPlatformSelected) platformIds.push(platform.id)
    })

    const newData = {
      ...data,
      jobs: {
        connect: jobIds,
      },
      platforms: {
        connect: platformIds,
      },
    }

    onSubmitHandler(newData as any)
  }
  ```

Comment (MustafaTRKYLMZ) [reply packages/ui/src/components/JoinForm/JoinForm.tsx:70]: First part is working. But I didn't tried second part yet.

Comment (7alip) [reply packages/ui/src/components/JoinForm/JoinForm.tsx:70]: I mean, it creates a profile data but without jobs and platforms.

Another thing I would mention is: Should we really save `platforms` information. We can already get this data from `const profilePlatforms = profile.jobs.map(job => job.platform)`

File: packages/ui/src/data/requestArgs/index.ts [hunk 3]
```diff
@@ -78,7 +86,7 @@ export const useRequestArgs = (): UseRequestArgsReturn => {
           operator: '$eq',
         },
       ],
-      populate: ['users.role', 'platforms'],
+      populate: ['users.role', 'platforms', 'jobs', 'role'],
```
7alip, line 89: If we will remove `platforms` relation from profile, we can still reach the platform data with nested populate

```suggestion
      populate: ['users.role', 'jobs.platform', 'role'],
```

packages/ui/src/data/tables/profile.tsx
```diff
@@ -54,11 +64,17 @@ export const useProfileColumns = (): WTableProps<
...
+        (value as Platform[])?.map(p => p[`name_${locale}`]).join(', '),
+      sortable: true,
+      sortKey: `slug`,
+    },
+    jobs: {
+      transform: value =>
+        (value as Job[])?.map(job => job[`name_${locale}`]).join(', '),
       sortable: true,
       sortKey: `slug`,
     },
```
7alip, line 77: If we remove `platforms` relation from profile model, we can't show `jobs` and `platforms` at the same time on the table. Because we can show `platforms` through `jobs` field. 

We need to change WTable implementation to fix this kind of issues. I'll try to change columns from object to array to allow displaying the same field multiple times

PR #1010, opened 2023-12-19
Update TweetGenAi

7alip:


packages/ui/src/components/PostSentenceForm/PostSentenceForm.tsx
```diff
@@ -39,7 +40,8 @@ export const PostSentenceForm: FC<PostSentenceFormProps> = ({
   }
 
   return (
-    <Stack p={4} spacing={4}>
+    <Stack spacing={4}>
+      {/* TODO: Move this into a separate PostSentenceCreator component */}
```
7alip, line 44: @hakanayata Could you create a new task for this?

The new PostSentenceCreator component (with handleAdd) can be used in TweetGenAI component.

```ts
// Redis key format: <content>::<postId>::<shareCount>::<published>
export type RedisPost = `${string}::${number}::${number}::${0 | 1}`
```

7alip: #1011

PR #1021, opened 2023-12-24
895 implement recaptcha for blog like and view

hasancomce:


apps/api/src/api/blog/controllers/custom.ts
```diff
@@ -43,4 +45,108 @@ export default {
 
     return { data: user }
   },
+  async like(ctx: Context) {
+    // await checkRecaptcha(ctx)
```
7alip, line 49: Can you enable this line, so that we can test it in production

apps/api/src/api/blog/controllers/custom.ts
```diff
@@ -43,4 +45,108 @@ export default {
...
+
+    const result = await strapi.entityService.findOne(
+      'api::blog.blog',
+      ctx.params.id,
+    )
+
+    return { data: result }
+  },
+  async unlike(ctx: Context) {
+    // await checkRecaptcha(ctx)
```
7alip, line 89: Enable this also

apps/api/src/api/blog/controllers/custom.ts
```diff
@@ -43,4 +45,108 @@ export default {
...
+    const result = await strapi.entityService.findOne(
+      'api::blog.blog',
+      ctx.params.id,
+    )
+
+    return { data: result }
+  },
+  async view(ctx: Context) {
+    try {
+      // await checkRecaptcha(ctx)
```
7alip, line 130: Enable this one

packages/services/src/art/like.ts
```diff
@@ -23,7 +22,11 @@ const useLikeArtMutation = () => {
   })
 }
 
-export const useLikeArt = (art?: Art | null) => {
+export const useLikeArt = (art?: Art | null, queryKey?: QueryKey) => {
```
7alip, line 25: `Type error: Cannot find name 'QueryKey'`. This part is causing build error.

We can remove `queryKey`. We can implement `onSuccess` later in a separate task

7alip (changes requested):
Great work, thanks. Could you fix the build problem.

apps/api/src/api/blog/controllers/custom.ts
hasancomce: I didn't notice :)

packages/services/src/art/like.ts
hasancomce: I did as you mentioned. Thanks.

apps/api/src/api/art/controllers/custom.ts
```diff
@@ -34,21 +34,15 @@ export default {
...
-        ctx.params.id,
-        {
+      if (!isLikedCount) {
+        await strapi.entityService.update('api::art.art', ctx.params.id, {
           data: {
             likers: {
-              [connection]: [profile.id as number],
-            } as any,
+              ['connect']: [profile.id as number],
+            },
```
7alip, line 42: I removed `as any`, seems Strapi TypeScript supports it now.

apps/api/src/api/art/controllers/custom.ts
```diff
@@ -66,6 +60,27 @@ export default {
   async unlike(ctx: Context) {
     await checkRecaptcha(ctx)
 
+    const profile = await getProfile(ctx)
+
+    if (profile) {
+      const isLikedCount = await strapi.entityService.count('api::art.art', {
+        filters: {
+          id: { $eq: ctx.params.id },
+          likers: { id: { $in: [profile.id] } },
```
7alip, line 69: It seems there was some misunderstanding about the use of the `$in` operator here. Ideally, we can use the `$in` operator to check if multiple provided ids are in the `likers` relation. However, we will always provide a single profile id. If `$eq` works as well, I'd use it. It could also be better for performance.

```suggestion
          likers: { id: { $eq: profile.id } },
```

apps/api/src/api/blog/controllers/custom.ts
```diff
@@ -43,4 +50,100 @@ export default {
...
+  async like(ctx: Context) {
+    await checkRecaptcha(ctx)
+
+    const profile = await getProfile(ctx)
+
+    if (profile) {
+      const isLikedCount = await strapi.entityService.count('api::blog.blog', {
+        filters: {
+          id: { $eq: ctx.params.id },
+          likers: { id: { $in: [profile.id] } },
```
7alip, line 62: If `$eq` works, this can be updated too.

apps/api/src/api/blog/controllers/custom.ts
```diff
@@ -43,4 +50,100 @@ export default {
...
+  async unlike(ctx: Context) {
+    await checkRecaptcha(ctx)
+
+    const profile = await getProfile(ctx)
+
+    if (profile) {
+      const isLikedCount = await strapi.entityService.count('api::blog.blog', {
+        filters: {
+          id: { $eq: ctx.params.id },
+          likers: { id: { $in: [profile.id] } },
```
7alip, line 98: The same comment...

7alip (commented):
I formatted code by running `yarn format`. You can always check if it will be built by running `yarn build:dev`. We haven't implemented type check script yet.

apps/api/src/api/art/controllers/custom.ts
hasancomce: Thanks. This part used to be like this. I didn't want to change it because I was new to TS types.

7alip (approved):
Thanks, do you want to merge it?

PR #1026, opened 2023-12-27
Burak - Responsive Card

7alip:


packages/ui/src/examples/E0_ResponsiveCard.tsx
```diff
@@ -1,31 +1,55 @@
...
+      h: '100%',
+      justifyContent: 'start',
+      alignItems: 'start'
+    },
+    image: {
+      w: '300px',
+      h: '100%',
+      objectFit: 'cover'
+    }
+  };
```
7alip, line 19: This is an alternative approach, perhaps similar to Material UI, that we have never used in the project. It will make the component markup more clean and visible, but you will lose the TypeScript autocomplete feature.

What I like about inline styling (tailwind, chakra ui, and many others) is that you can concentrate on the element you want to style. You are not required to scroll or switch between files.

packages/ui/src/examples/E0_ResponsiveCard.tsx
```diff
@@ -1,31 +1,55 @@
...
+
+  const verticalBuild = {
+    baseFlex: {
+      flexDirection: 'column',
+      w: '347px',
+      h: '707px',
```
7alip, line 25: These arbitrary constant dimensions are not good responsiveness practice. You could still use maxW or maxH props to limit it. I would recommend using `aspectRatio` for a responsive circle square image.

You can use `aspectRatio` prop or `<AspectRatio>` component

https://chakra-ui.com/docs/components/aspect-ratio/usage


```tsx
<Box aspectRatio={1} maxW={350}>
 <Image boxSize={'full'} /> // width="100%" height="100%"
</Box> 
```

packages/ui/src/examples/E0_ResponsiveCard.tsx
```diff
@@ -1,31 +1,55 @@
...
+    image: {
+      boxSize: '200px',
+      borderRadius: 'full'
+    }
+  }
+
+  const props = useBreakpointValue({
+    sm: horizontalBuild,
+    base: verticalBuild,
+  });
```
7alip, line 38: I used `useBreakpointValue` a lot in the components. But with Next.js, I try to  avoid using it for conditional rendering or styling. This can cause *hydration* problems. The hydration problem occurs when the HTML on the client and the server differ.

More info: https://chakra-ui.com/docs/hooks/use-breakpoint-value#usage

> This hook is built to work in server-side rendering (SSR) applications by default. You might notice a quick flash of incorrect media query values when you use them.
>
> If you're creating a client-side rendered app, you can leverage the ssr argument to get the correct value on the first render.
>
> ```tsx
> const buttonSize = useBreakpointValue({ base: 'sm', lg: 'md' }, { ssr: false })
>
> const breakpoint = useBreakpoint({ ssr: false })
> ```

7alip (commented):
Thanks. I left some feedback. You don't need to feel obligated to apply the changes. Onboarding tasks are just getting started of our stack.

7alip:
@burakndrlnd If you have no intention of pushing anything, we can close this PR.

PR #1030, opened 2023-12-31
Bbluechip responsive card

bbluechip:
ResponsiveCard.tsx dizaynları eklendi fakat ResponsiveCard.stories.tsx belgesi bir takım gözlemlemelere ihtiyaç duyabilir.


packages/ui/src/components/ResponsiveCard/ResponsiveCard.tsx
```diff
@@ -0,0 +1,74 @@
+import { Flex, Image, Stack, Heading, Text } from "@chakra-ui/react";
+
+const ResponsiveCard = () => {
+  const randomPic =
+    "https://upload.wikimedia.org/wikipedia/en/3/34/Blue_%28film%29.png";
+
+  return (
+    <Flex
+      margin="50px auto"
+      boxShadow="0 4px 8px rgba(0, 0, 0, 0.1)"
```
7alip, line 10: Component itself should not have margins by default. If you want to adjust the alignment of the component on the story page, you could apply margin to story decorator in the config.

```tsx
decorators: [Story => <Container maxW="container.sm" m="50px auto">{Story()}</Container>],
```

packages/ui/src/components/ResponsiveCard/ResponsiveCard.tsx
```diff
@@ -0,0 +1,74 @@
...
+      flexDirection="row"
+    >
+      <Image
+        src={randomPic}
+        alt="random pic"
+        objectFit="cover"
+        w={{ base: "auto", md: "50%" }}
+      />
+
+      <Stack margin="20px">
```
7alip, line 20: I would recommend using `padding` instead.

packages/ui/src/components/ResponsiveCard/ResponsiveCard.tsx
```diff
@@ -0,0 +1,74 @@
...
+};
+
+const ResponsiveCardColumn = () => {
+  const randomPic =
+    "https://upload.wikimedia.org/wikipedia/en/3/34/Blue_%28film%29.png";
+
+  return (
+    <Flex
+      margin="50px auto"
+      boxShadow="0 4px 8px rgba(0, 0, 0, 0.1)"
```
7alip, line 44: Remember that you can use theme settings like `boxShadow="sm"` here. It's most likely using CSS variables to improve consistency. You can always update or add your own styles to the theme.

packages/ui/src/components/ResponsiveCard/ResponsiveCard.tsx
```diff
@@ -0,0 +1,74 @@
...
+      flexDirection="column"
+      align="center"
+      maxW="347px"
```
7alip, line 47: I wouldn't limit the width of the wrapper element which breaks responsiveness. The component will not fill it's container on most larger mobile devices.

packages/ui/src/components/ResponsiveCard/ResponsiveCard.tsx
```diff
@@ -0,0 +1,74 @@
...
+      <Stack position="relative" overflow="hidden" borderRadius="50%" mt={5}>
+        <Image
+          borderRadius="full"
+          boxSize="200px"
+          src={randomPic}
+          alt="Dan Abramov"
+        />
+      </Stack>
+
+      <Stack p="20px" w="100%">
```
7alip, line 59: Nice that you use padding here. Remember that there are pre-configured dimensions that you can use for width, height, padding, margin etc. Instead of pixels, I prefer to use these values. For example, `p={4} mt={6} h={16}`.

7alip (commented):
Thank you for your submission.

The task was probably misunderstood. I was expecting a single responsive component instead of two separate components. Component design should adapt to viewport width.

You can read [my comment](https://github.com/wsvvrijheid/monorepo/pull/47#discussion_r944140256) again which was the task description.

PR #1031, opened 2024-01-02
onboarding/fetch-bbluechip (FetchWithStrapiRequest)

bbluechip:
Changed ResponsiveCard design.

### DIFF
--- a/packages/ui/src/examples/E0_ResponsiveCard.tsx
+++ b/packages/ui/src/examples/E0_ResponsiveCard.tsx
@@ -1,33 +1,63 @@
-import { Flex, Image, Stack, Heading, Text } from '@chakra-ui/react'
+import {
+  Flex,
+  Image,
+  Stack,
+  Heading,
+  Text,
+  useBreakpointValue,
+} from "@chakra-ui/react";
 
 export const ResponsiveCard = () => {
-  const randomPic = 'https://picsum.photos/600/400'
+  const randomPic =
+    "https://www.formica.com/-/media/project/formica/emea/products/swatch-images/f2828/f2828-swatch.jpg";
+  const isWideVersion = useBreakpointValue({ base: false, md: true });
+  
+  return isWideVersion ? (
+    <Flex boxShadow="0 4px 8px rgba(0, 0, 0, 0.1)" flexDirection="row">
+      <Image src={randomPic} alt="random pic" objectFit="cover" width="250px" />
 
-  return (
-    <Flex>
-      <Stack>
-        <Image src={randomPic} alt="random pic" />
-      </Stack>
-
-      <Stack>
-        <Heading as="h1" size="lg" mb={5} color="#000" fontWeight="500">
-          This is Heading...
+      <Stack p={4}>
+        <Heading as="h1" size="lg" mb={5} color="#000" fontWeight="600">
+          Lorem ipsum dolor sit amet consectetur adipisicing elit. Esse?
         </Heading>
         <Text>
-          Lorem ipsum dolor sit amet consectetur adipisicing elit. Commodi
-          dolores totam blanditiis soluta quae delectus doloremque velit sit
-          corrupti iure natus nobis nesciunt, tempore repellat cupiditate culpa
-          voluptatum expedita. Ducimus sapiente exercitationem repellendus et
-          aspernatur. Aspernatur expedita a ea non iure aperiam assumenda
-          praesentium beatae numquam repellat, excepturi impedit voluptatem!
-          Lorem ipsum dolor sit amet consectetur adipisicing elit. Commodi
-          dolores totam blanditiis soluta quae delectus doloremque velit sit
-          corrupti iure natus nobis nesciunt, tempore repellat cupiditate culpa
-          voluptatum expedita. Ducimus sapiente exercitationem repellendus et
-          aspernatur. Aspernatur expedita a ea non iure aperiam assumenda
-          praesentium beatae numquam repellat, excepturi impedit voluptatem!
+          Lorem ipsum dolor sit amet, consectetur adipisicing elit. Alias
+          veritatis quidem voluptatibus nihil aperiam adipisci harum, labore
+          perferendis. Sunt reiciendis dignissimos iure ut enim reprehenderit
+          ratione ipsam officiis deserunt consectetur quam consequatur corporis
+          pariatur facilis impedit magnam explicabo necessitatibus, sit vel
+          tempora nam doloribus nesciunt cumque ex? Alias, aperiam placeat?
         </Text>
       </Stack>
     </Flex>
-  )
-}
+  ) : (
+    <Flex
+      margin="50px auto"
+      boxShadow="sm"
+      flexDirection="column"
+      align="center"
+      w="70%"
+    >
+      <Stack position="relative" overflow="hidden" borderRadius="50%" mt={5}>
+        <Image
+          borderRadius="full"
+          boxSize="200px"
+          src={randomPic}
+          alt="random picture"
+        />
+      </Stack>
+
+      <Stack p={4} w="100%">
+        <Heading as="h1" size="lg" mb={5} color="#000" fontWeight="600">
+          Lorem ipsum dolor sit amet consectetur adipisicing elit. Esse?
+        </Heading>
+        <Text>
+          Lorem ipsum dolor sit, amet consectetur adipisicing elit. Natus a
+          expedita iure, doloremque corporis nemo sint repellat voluptas
+          sapiente eum assumenda minima earum officiis ut atque. Tenetur vel
+          culpa exercitationem.
+        </Text>
+      </Stack>
+    </Flex>
+  );
+};
--- a/packages/ui/src/examples/E1_FetchWithAxios.tsx
+++ b/packages/ui/src/examples/E1_FetchWithAxios.tsx
@@ -19,7 +19,7 @@ export const FetcWithAxios = () => {
     const fetchData = async () => {
       try {
         const response = await axios.get(BLOG_URL)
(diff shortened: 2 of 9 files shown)


packages/ui/src/examples/E0_ResponsiveCard.tsx
```diff
@@ -1,33 +1,63 @@
...
+  Heading,
+  Text,
+  useBreakpointValue,
+} from "@chakra-ui/react";
 
 export const ResponsiveCard = () => {
-  const randomPic = 'https://picsum.photos/600/400'
+  const randomPic =
+    "https://www.formica.com/-/media/project/formica/emea/products/swatch-images/f2828/f2828-swatch.jpg";
+  const isWideVersion = useBreakpointValue({ base: false, md: true });
```
7alip, line 13: `useBreakpointValue` uses JavaScript to calculate the viewport width under the hood. We focus on using pure css approach when making our components responsive.

For example, in plain css we use [media queries.](https://www.w3schools.com/css/css3_mediaqueries.asp). ChakraUI and some other competitors try to simplify using media queries.

**With ChakraUI**
```tsx
<Box width={{ base: "full", md: "50%", lg: "25%" }} /> // Object notation
// or
<Box width={["full", "full", "50%", "25%"] /> // Array notation
```

**With TailwindCSS**
```tsx
<div className="w-full md:w-1/2 lg:w-1/4"  />
```

**With plain css**
```css
div {
  width: 100%;
}

@media screen and (min-width: 30em) {
  div {
    width: 50%;
  }
}

@media screen and (min-width: 48em) {
  div {
    width: 25%;
  }
}
```

> Remember that the **Mobile First** approach uses `(min-width)` in the queries, while the **Desktop First** approach uses `(max-width)`. And most library uses mobile first approach.

7alip (commented):
Thanks. I left a comment about your solution.

packages/ui/src/examples/E1_FetchWithAxios.tsx
```diff
@@ -2,24 +2,45 @@
...
 import { API_URL } from '@wsvvrijheid/config'
 
 // https://wsvv-api-staging.onrender.com/api/blogs?locale=tr
 // You can use local API_URL instead of the above url
 // yarn --cwd apps/api dev to run the api locally (http://localhost:1337)
-const BLOG_URL = `${API_URL}/api/blogs`
+const BLOG_URL = "https://wsvv-api-staging.onrender.com/api/blogs?locale=tr"
 
 export const FetcWithAxios = () => {
   const [blogs, setBlogs] = useState([])
```
7alip, line 16:
```suggestion
  const [blogs, setBlogs] = useState<Blog[]>([])
```

You can specify the type of the state. It helps.

packages/ui/src/examples/E1_FetchWithAxios.tsx
```diff
@@ -2,24 +2,45 @@
...
-  return <Box>{/* TODO: Show only title of the blogs */}</Box>
+  return
+   (
+  <Box>
+    {blogs.map((blog, index) => (
+    <div key={index}>{blog}</div>
+  ))}
+  <p>hello</p>
+  </Box>
+  )
```
7alip, line 45: It seems your code is not formatted. You can check if you have [EsLint extension](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) installed in VSCode. I recommend using VSCode. If you have any other IDE, you can investigate how to format your code based on the [prettier config](https://github.com/wsvvrijheid/monorepo/blob/main/.prettierrc) in the root of the project.

7alip (approved):
Thanks

packages/ui/src/examples/E1_FetchWithAxios.tsx
```diff
@@ -2,24 +2,45 @@
...
-  }, [])
+    fetchBlogs();
+  }, [locale])
 
-  return <Box>{/* TODO: Show only title of the blogs */}</Box>
+  return
+   (
+  <Box>
+    {blogs.map((blog, index) => (
+    <div key={index}>{blog}</div>
```
7alip, line 41: `{blog}` is not renderable by React because it's an object. 

```suggestion
    <div key={index}>{blog.title}</div>
```

7alip (changes requested):

packages/ui/src/examples/E1_FetchWithAxios.tsx
```diff
@@ -2,24 +2,45 @@
...
+
     // TODO: fetch blogs with axios by using the API_URL and TOKEN (Authorization header with Bearer token)
     // NOTE: Not every locale may have a blog, so you may need to change the locale to tr
     // REF: https://docs.strapi.io/dev-docs/plugins/i18n#getting-localized-entries-with-the-locale-parameter
-  }, [])
+    fetchBlogs();
+  }, [locale])
 
-  return <Box>{/* TODO: Show only title of the blogs */}</Box>
+  return
```
7alip, line 37: @bbluechip You won't see the results in Storybook, the reason is that you don't return the components but undefined because of unformatted code.

Please right click on the code and select `Format Document`. Make sure you have Prettier and ESlint extensions. When you format your code, you will notice that the JSX part of your component is unreachable.

| Original (unformatted)                                                                                 | Formatted                                                                                              | Correct                                                                                                |
| ------------------------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------ | ------------------------------------------------------------------------------------------------------ |
| image | image | image |

7alip: You can also see that the JSX has pale colors which means this part is not reachable because of the early return.

Screenshot 2024-01-04 at 03 24 19

packages/ui/src/examples/E1_FetchWithAxios.tsx
```diff
@@ -1,25 +1,33 @@
...
+    fetchData()
   }, [])
 
-  return <Box>{/* TODO: Show only title of the blogs */}</Box>
-}
+  return (
+    <Box>
+      <ul>
+        {blogs.map((blog) => (
+          <li key={blog.id}>{blog.title}</li>
```
7alip, line 28: TypeScript probably complains about `blog.id` and `blog.title`, because blog type is not specified. You can indicate the blog state type as following:

```tsx
const [blogs, setBlogs] = useState<Blog[]>([])
```

bbluechip:
Added the interface of the Blog for not returning an error in useState line, need to specify but not defining a variable. 

Added response.data.data => First I thought respose.data would be enough but then I remembered, response.data is just the base returning data. Because of the api has data, meta objects. I need to write one more .data after response.data 

In this version FetchWithAxios running well.

7alip (approved):
Thanks. You can close the task now :)

packages/ui/src/examples/E2_FetchWithStrapiRequest.tsx
```diff
@@ -1,17 +1,51 @@
...
+  const [blogs, setBlogs] = useState<Blog[]>([]);
 
   useEffect(() => {
-    // TODO: fetch blogs with strapiRequest by using the API_URL and PUBLIC_TOKEN
-    // Remember that fetcher is a wrapper around axios that adds the token and api url to the request
-  }, [])
+    const fetchData = async () => {
+      try {
+        // I couldn't find any argument for api url in strapiRequest function
+        const response = await strapiRequest({
```
7alip, line 23: As the name suggests, `strapiRequest` is simply an axios wrapper for fetching data from our Strapi backend. It's a generic function (`function<T> {}`) that allows you to specify which type of Strapi model _(Blog, Activity, Hashtag etch)_ you want to retrieve data from.

```suggestion
        const response = await strapiRequest<Blog>({
```

packages/ui/src/examples/E2_FetchWithStrapiRequest.tsx
```diff
@@ -1,17 +1,51 @@
...
+          endpoint: 'blogs',
+          token: PUBLIC_TOKEN,
```
7alip, line 25: You should specify the `locale` for some models, otherwise only the models in the default language (en) will be retrieved.

```suggestion
          locale: 'tr' // or preferably locale from useRouter()
          token: PUBLIC_TOKEN,
```

packages/ui/src/examples/E2_FetchWithStrapiRequest.tsx
```diff
@@ -1,17 +1,51 @@
...
+        });
+        // There is no response data here. 
+        console.log(response)
 
-  return <Box>{/* TODO: Show only title of the blogs */}</Box>
-}
+        // property data does not exist on type 'StrapiModel[] so I write an if statement here 
+        if ('data' in response.data)
+          setBlogs(response.data.data as Blog[])
```
7alip, line 32: This statement is not necessary because `strapiRequest` should return as the following format: 
`{ data: [], pagination: {} }`


There are some other advanced debugging options, such as Network tab.

You can check to which endpoint are you making a request, what is your payload and what the server responses.

In your case, you are probably making a request to `http://localhost:1337`, which might not be running locally.

**You have some options to make it work:**

1. Update your env  `NEXT_PUBLIC_API_URL=http://localhost:1337` to`NEXT_PUBLIC_API_URL=https://wsvv-api-staging.onrender.com` in **packages/ui/.env**
2. Run your local server  by adding `"storybook": "strapi dev"` to your `apps/api/package.json` scripts.
3. Run your local server by running `cd apps/api && yarn dev`
4. If you are unable fix with these options, you can statically change API_URL in your strapiRequest function

Screenshot 2024-01-05 at 22 31 31

7alip (commented):
Thanks, I think you should just update your env variables or run your local server to make it work.

PR #1035, opened 2024-01-06
Burak - Onboarding 3 - Category

7alip:


packages/ui/src/examples/E6_CreateCategory.tsx
```diff
@@ -29,12 +30,19 @@ export const CreateCategory: FC<CreateCategoryWithAxiosProps> = ({
...
+      name_en,
+      name_tr,
+      name_nl,
+      slug,
+    }
 
     if (fetcher === 'axios') {
-      // const response = await createCategoryWithAxios(categoryBody)
-      // blogs = response.data
+      const response = await createCategoryWithAxios(categoryBody)
```
7alip, line 42: It seems we need to update `createCategoryWithAxios` function. Consider adding PUBLIC_TOKEN if it exists. Otherwise the header can be `Authorization: 'Bearer undefined'` which will always result 401 error.

```ts
export const createCategoryWithAxios = async (data: CategoryCreateInput) => {
  return await axios.post(
    CATEGORY_URL,
    { data },
    {
      ...(PUBLIC_TOKEN && {
        headers: { Authorization: `Bearer ${PUBLIC_TOKEN}` },
      }),
    },
  )
}
```

7alip (commented):
I changed the category POST endpoint to accept requests without a token. It should be working now. The rest looks perfect, thanks.

PR #1036, opened 2024-01-07
create responsive card

ermissa:


packages/ui/src/examples/E0_ResponsiveCard.tsx
```diff
@@ -4,28 +4,42 @@ export const ResponsiveCard = () => {
   const randomPic = 'https://picsum.photos/600/400'
 
   return (
-    <Flex>
-      <Stack>
-        <Image src={randomPic} alt="random pic" />
+    <Flex shadow="lg" direction={{ base: 'column', sm: 'row' }}>
```
7alip, line 7: Just a quick note that I would prefer to use lg, or at the very least md breakpoints instead. The rest is perfect, thanks.

```suggestion
    <Flex shadow="lg" direction={{ base: 'column', lg: 'row' }}>
```

7alip (approved):
Thanks, you can close the PR

PR #1037, opened 2024-01-07
Remove saved AI-generated posts from the list

hakanayata:
Remove AI-generated posts from the stack once they are added to the sentences section

### DIFF
--- a/packages/ui/src/components/PostSentenceCreator/PostSentenceCreator.tsx
+++ b/packages/ui/src/components/PostSentenceCreator/PostSentenceCreator.tsx
@@ -1,4 +1,4 @@
-import { useState } from 'react'
+import { useEffect, useState } from 'react'
 
 import { HStack, IconButton, Textarea, ThemeTypings } from '@chakra-ui/react'
 import { useQueryClient } from '@tanstack/react-query'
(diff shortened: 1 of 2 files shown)


packages/ui/src/components/PostSentenceCreator/PostSentenceCreator.tsx
```diff
@@ -11,29 +11,38 @@ type PostSentenceCreatorProps = {
...
     onAddMutation.mutate(
       { hashtagId, value: `${value}::${postId}::${0}::${0}` },
       {
-        onSuccess: () =>
+        onSuccess: () => {
           queryClient.invalidateQueries({
             queryKey: ['kv-hashtag-sentences', hashtagId],
-          }),
+          })
+          onRemove && onRemove()
```
7alip, line 42: Isn't it more meaningful to use the name `onSuccess` as the prop name? I know you use that callback to remove the item, but it's actually called after the mutation is successful.

Remember you can also use the [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining) for functions as well. `onRemove?.()`

7alip (approved):
Thanks

7alip (approved):
Thanks, you can merge it

PR #1038, opened 2024-01-08
Provide the blog page with recaptcha provider

hasancomce:


apps/foundation/pages/blog/[slug].tsx
```diff
@@ -42,18 +43,20 @@ const BlogDetailPage: FC<BlogPageProps> = ({
...
-          post={blog}
-          source={source}
-          link={link}
-          isLiked={isLiked as boolean}
-          toggleLike={toggleLike}
-          authorBlogs={authorBlogs}
-        />
-      </Container>
-    </Layout>
+    <ReCaptchaProvider reCaptchaKey={RECAPTCHA_SITE_KEY}>
```
Comment (7alip) [inline apps/foundation/pages/blog/[slug].tsx:46]: `useRecaptchaToken` and `useLikeBlog` should be used in RecaptchaProvider. In this case, we call `useLikeBlog` on the `[slug]` page, outside of the provider.  That's why we can't get the recaptcha token, I suppose.

Comment (7alip) [reply apps/foundation/pages/blog/[slug].tsx:46]: @hasancomce You can try using `useLikeBlog` in `BlogDetail`

Comment (7alip) [reply apps/foundation/pages/blog/[slug].tsx:46]: Alternatively you can use higher order component:

```tsx
import React from 'react';
import { ReCaptchaProvider } from 'react-recaptcha-x';

// This is your HOC
const withReCaptcha = (WrappedComponent) => {
  return (props) => (
    <ReCaptchaProvider siteKey="your-site-key">
      <WrappedComponent {...props} />
    </ReCaptchaProvider>
  );
};

export default withReCaptcha
```

Then

```tsx
export default withReCaptcha(BlogDetailPage)
```

7alip (commented):
Thanks

7alip (approved):
Thanks, you can merge it

PR #1053, opened 2024-02-04
Fix locale based fetching of recommended tweets/topics

ertugrul-aydin:
Fixes tweet/topic fetching based on user language. Content now updates correctly when language changes.


7alip (approved):
Thanks, it's simpler than I thought :)

PR #1058, opened 2024-02-06
added note field in CourseApplicationFields and CourseApplicationSchema

bbluechip:
I added notes field on strapi admin  and then I edit the courseapplicationfield and courseapplicationschema


7alip (changes requested):
Can you do the followings:

- `yarn format`
- Add `notes` to `packages/ui/src/components/CourseDetailPage/schema.ts`

7alip:
@bbluechip I think we shouldn't include `notes` in `CourseApplicationForm`. `notes` will be used by admins/editors in Dashboard website. We use `CourseApplicationForm` in the foundation website for users.

You can make `notes` field optional so that TypeScript won't give an error in CourseApplicationForm

```ts
// packages/types/src/course-application.ts

export type CourseApplicationCreateInput = Omit<
  CourseApplicationBase,
  'hasPaid' | 'approvalStatus' | 'notes'
> & {
  course: number
  notes?: string
}
```

PR #1073, opened 2024-02-28
Create new custom router endpoint for profile/me

bbluechip:
needen review. Dealing some problem while trying to fetch profile data using getSessionUser function could't able to import file.

### DIFF
--- a/apps/api/src/api/profile/router/custom.ts
+++ b/apps/api/src/api/profile/router/custom.ts
@@ -1,0 +1,28 @@
+// apps/api/src/api/profile/router/custom.ts
+
+import { NextApiRequest, NextApiResponse } from 'next';
+
+// I get an error while I have trying to fetch profile information, am I using the right function? I cannot import this function. 
+import { getSessionUser } from '../../../../../../packages/services/src/auth/getSessionUser';
+
+export default async function handler(req: NextApiRequest, res: NextApiResponse) {
+  if (req.method !== 'GET') {
+    res.status(405).json({ message: 'Method Not Allowed' });
+    return;
+  }
+
+  try {
+    const userToken = 'burayaKullanıcınınTokenBilgisi';
+    const user = await getSessionUser(userToken);
+
+    const profile = await user; 
+
+    
+    const singleResponse = Array.isArray(profile) ? profile[0] : profile;
+
+    res.status(200).json(singleResponse);
+  } catch (error) {
+    console.error(error);
+    res.status(error.statusCode || 500).json({ message: error.message || 'Internal Server Error' });
+  }
+}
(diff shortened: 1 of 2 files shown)


7alip (changes requested):
I put some comments to clarify things.

You need to focus on Strapi backend part (routes and controllers), but it seems you've tried to implement Nextjs API handler to Strapi backend.

See other endpoint examples such as topic, hashtag etc. We use custom routes and controllers in some other endpoints too.

bbluechip:
How can I add profiles/me endpoint to strapiSingleEndpoint type,
How can I update profile request with the new endpoint after login.

7alip:
@bbluechip 

> How can I add profiles/me endpoint to strapiSingleEndpoint type, How can I update profile request with the new endpoint after login.

- Search for `StrapiSingleEndpoint` in VSCode. `(packages/types/src/strapi.ts)`

---

- Search for `profileResponse` (I mentioned in the task for getAuth but the following files can be adjusted)
  - packages/services/src/api/reset-password.ts
  - packages/services/src/api/user.ts
  - packages/services/src/auth/getAuth.ts

apps/api/src/api/profile/controllers/custom.ts
```diff
@@ -0,0 +1,27 @@
...
+    async getProfile(ctx) {
+      if (!ctx.state.user) {
+        throw ctx.unauthorized('You are not authenticated');
+      }
+  
+      const userId = ctx.state.user.id;
+  
+      try {
+        // what should I use here? 
+        const profile = await strapi.entityService.findOne('api::profile.profile', userId, {
```
7alip, line 13: You can't find the correct profile by using `userId`. It would work if you use `profileId` here. Since Profiles and Users tables are not the same, the ids are also not the same.

Check `getProfile` function in `apps/api/src/utils/getProfile.ts`. There we find the profile with `userId` like the following:

```ts
const profileResponse = await strapi.entityService.findMany(
    'api::profile.profile',
    {
      filters: {
        user: { id: { $eq: ctx.state.user.id } },
      },
    },
  )

  const profile = profileResponse?.[0] || null
```

You can add populate here.

7alip (changes requested):
I left a comment in the profile controllers. We're probably done after this small adjustment.

7alip:
@bbluechip I merged dev into this branch and removed `[id] 2.tsx` page. Please pull it and run `yarn install` first.

apps/api/src/api/profile/controllers/custom.ts
7alip: @bbluechip There were probably misunderstanding. I didn't expect your branch to be merged because 

```ts
const profile = await strapi.entityService.findOne('api::profile.profile', userId, {
  populate: 'user', 
});
```

this part should be replaced with:

```ts
const profileResponse = await strapi.entityService.findMany(
  'api::profile.profile',
  {
    filters: {
      user: { id: { $eq: ctx.state.user.id } },
    },
  },
)

const profile = profileResponse?.[0] || null
```

Would you like to fix this in a new branch?

Issue #1024, opened 2023-12-25
Blog like/unlike request not working properly


hasancomce:
When the like/unlike button is clicked repeatedly, it sends the same request. It should work in the `like/unlike` loop. 

According to the response after the blog like/unlike request, the blog in the context may not be updated. 

Request 1: 
<img width="841" alt="1" src="https://github.com/wsvvrijheid/monorepo/assets/79446497/a9a950b9-209e-4fae-aa54-fd045e60cbbe">

Request 2: 
<img width="841" alt="2" src="https://github.com/wsvvrijheid/monorepo/assets/79446497/90724ded-a312-4eab-a7c5-63238b688d28">

Request 3: 
<img width="841" alt="3" src="https://github.com/wsvvrijheid/monorepo/assets/79446497/3cbc1773-75b3-4de9-8296-fd7001574a80">

burakndrlnd:
![image](https://github.com/freedomcombination/monorepo/assets/154612391/f7736bd0-9582-4f62-a442-650af0b3f7f8)

@7alip this is how console look like after i click like button. im unregistered user. i think public user role has no permission those endpoints.

burakndrlnd:
I was not running it on local. It was freedomcombination.com

7alip:
> I was not running it on local. It was freedomcombination.com

If you click the button repeatedly, recaptcha validation may reject your request. We implemented this intentionally to prevent bots from liking or disliking the content.

PR #1090, opened 2024-03-13
blog like button.

burakndrlnd:
screen

https://github.com/freedomcombination/monorepo/issues/1024#issuecomment-1995685976

### DIFF
--- a/packages/services/src/blog/like.ts
+++ b/packages/services/src/blog/like.ts
@@ -1,12 +1,13 @@
 import { QueryKey, useMutation, useQueryClient } from '@tanstack/react-query'
 import axios from 'axios'
-import { useLocalStorage } from 'usehooks-ts'
+import { useDebounceValue, useLocalStorage } from 'usehooks-ts'
 
 import { API_URL } from '@fc/config'
 import { useAuthContext } from '@fc/context'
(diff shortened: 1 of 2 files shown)


7alip:
@burakndrlnd It triggers always unlike endpoint and never like. For example I click once and it successfully triggers `unlike`. Then I reload the page and click once more, it triggers `unlike` again.

packages/services/src/blog/like.ts
```diff
@@ -35,47 +36,65 @@ export const useLikeBlog = (blog?: Blog | null, queryKey?: QueryKey) => {
...
-          onSuccess: async () => {
-            await queryClient.invalidateQueries({ queryKey })
-
-            const updatedStorage = isLikedStorage
-              ? likersStorage?.filter(id => id !== blog.id)
-              : [...(likersStorage || []), blog.id]
-
-            setLikersStorage(updatedStorage as number[])
+        )
+      } else {
```
7alip, line 62: Since we early return from the first `if` statement, else is not necessary here. I overlooked this.

7alip:
@burakndrlnd You can close this and create a new issue as mentioned in https://github.com/freedomcombination/monorepo/pull/1091#issuecomment-1996338623 if you would like to.

PR #1091, opened 2024-03-14
Fix blog liking

7alip:
I noticed we were not using react-query's hydrated state properly. The blog data was always based on server side rendering and invalidating blog queryKey was not working.

Apart from that, we use localstorage in blog detail component. To prevent hydration issue, I exported the component dynamically without server rendering.

I also disabled the like button when recaptcha or authorization issue occurs.

### DIFF
--- a/apps/foundation/pages/blog/[slug].tsx
+++ b/apps/foundation/pages/blog/[slug].tsx
@@ -1,6 +1,6 @@
 import { FC } from 'react'
 
-import { QueryClient } from '@tanstack/react-query'
+import { QueryClient, dehydrate } from '@tanstack/react-query'
 import { GetServerSidePropsContext, InferGetServerSidePropsType } from 'next'
 import { useRouter } from 'next/router'
 import { serialize } from 'next-mdx-remote/serialize'
@@ -17,13 +17,7 @@ import { Layout } from '../../components'
 
 type BlogPageProps = InferGetServerSidePropsType<typeof getServerSideProps>
 
-const BlogDetailPage: FC<BlogPageProps> = ({
-  seo,
-  blog,
-  queryKey,
-  authorBlogs,
-  source,
-}) => {
+const BlogDetailPage: FC<BlogPageProps> = ({ seo, authorBlogs, source }) => {
   const {
     locale,
     query: { slug },
@@ -39,13 +33,7 @@ const BlogDetailPage: FC<BlogPageProps> = ({
     <ReCaptchaProvider reCaptchaKey={RECAPTCHA_SITE_KEY}>
       <Layout seo={seo}>
         <Container maxW="container.md">
-          <BlogDetail
-            post={blog}
-            queryKey={queryKey}
-            source={source}
-            link={link}
-            authorBlogs={authorBlogs}
-          />
+          <BlogDetail source={source} link={link} authorBlogs={authorBlogs} />
         </Container>
       </Layout>
     </ReCaptchaProvider>
@@ -81,11 +69,10 @@ export const getServerSideProps = async (
 
   return {
     props: {
-      blog,
       seo,
       source,
-      queryKey,
       authorBlogs,
+      dehydratedState: dehydrate(queryClient),
       ...(await ssrTranslations(locale)),
     },
   }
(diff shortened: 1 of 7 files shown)


burakndrlnd (approved):
i have `RECAPTCHA_SECRET_KEY=6LcBrGQ...rklS` in my .env file. but i endup with disabled like button.
beside that, instead of accumulate actions, you added isLoading state. my other pull request has become redundant :)

7alip:
> i have `RECAPTCHA_SECRET_KEY=6LcBrGQ...rklS` in my .env file. but i endup with disabled like button. beside that, instead of accumulate actions, you added isLoading state. my other pull request has become redundant :)

@burakndrlnd Sorry for taking your task but the main problem was hydrated state and using localstorage hook in server rendered component. I'll just wanted to show you some concepts, such as using prefecthing query with hydrated state.

We still have an issue, you can maybe work on a followup task:

In theory, a blog can have thousands of likers. What we are trying to do is that we populate likers field and check if the current user is in that likers array. Probably 100 profiles will be included at max. This is not the correct way of checking if a blog is liked by an authenticated user.

**Possible Solution:**

We should extend `blogs` find controller and add a new field `isLiked` if the request is made by a token. Or we can create a additional route for it. 

**Current Comparison:**

```ts
const isLikedByUser =
    profile &&
    blog?.likers &&
    blog?.likers?.length > 0 &&
    blog?.likers?.some(({ id }) => id === profile.id)
```

**Expected Comparison**
 
```ts
const isLikedByUser = blog.isLiked
```

Feel free to open a new issue if you would like to work on. The same approach should be implemented for arts too.

7alip:
@burakndrlnd You ended up with disabled state because recaptcha doesn't allow you to make multiple requests in a session. This doesn't allow you to unlike a blog unless you refresh the page. This doesn't look perfect, I know. With loading state, we maybe don't need a recaptcha validation, what do you think?

What do you mean by "instead of accumulate actions, you added isLoading state."?

burakndrlnd:
> @burakndrlnd You ended up with disabled state because recaptcha doesn't allow you to make multiple requests in a session. This doesn't allow you to unlike a blog unless you refresh the page. This doesn't look perfect, I know. 

i see. i didnt know that recaptcha prevent requests.

> With the help of the loading state, we maybe don't need a recaptcha validation (we may still need your debounced solution to prevent consecutive likes/dislikes), what do you think?
> What do you mean by "instead of accumulate actions, you added isLoading state."?

what i mean by accumulate actions is debounce solution.  it doesnt matter how many times you clicked. with 600ms delay, it compares your final request (like or not like) with initial request. if anything changes then send it to the server. but by the way you see all changes on client side.
by 'isLoading state' is your implementation. it waits user to complete last request then button activates. 

still both way recapcha ll prevent second request.

> @burakndrlnd Sorry for taking your task but the main problem was hydrated state and using localstorage hook in server rendered component. I'll just wanted to show you some concepts, such as using prefecthing query with hydrated state.

i dont mind. but thanks. i only read a few think about hydrate when i start this project. but understanding is getter wider by the time. after you send review request. i study it a little bit more. in teory, it sends an object from server side to client side and render dom as much as need not all, again in teory i think i get it. but in prantice, why is this needed i still cant figure it out yet. rendering something in server side in security reason i get it. but beside that ever server tick means money. and most of those works can easly and quickly be done in client side. i need brewing :)

> We should extend blogs find controller and add a new field isLiked if the request is made by a token. Or we can create a additional route for it.

for this solution, we still need to populate all liker or at least one more query to check likers includes user with that token. am i right?

7alip:
@burakndrlnd The debounce approach was wise, but I would have done it differently. Instead of using `useEffect` with useDebouncedValue, I would use `lodash.debounce`.

```ts
import { debounce } from 'lodash';

const toggleLike = debounce(async () => {}, 500)
```

---

Server side rendering is more for making blog pages more SEO friendly rather than for security reasons.

---

> for this solution, we still need to populate all liker or at least one more query to check likers includes user with that token. am i right?

Not actually, in server response there would be an additional field which is not in the database:

```ts
const blogs = [
  {
    id: 1,
    title: "Blog",
    ...,
    isLiked: true, // this field could be included from the server 
  }
]
```

PR #1093, opened 2024-03-14
Add GeneratedPostsContext to hashtag detail page

hakanayata:
This PR addresses the issue #1071


7alip:
@hakanayata Could you try fixing the conflict for `apps/dashboard/pages/hashtags/[id].tsx`?

packages/ui/src/post-maker/ArchivePostGenAI/GenPostProvider.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+export const useGenPostContext = () => {
+  return useContext(GenPostContext)
+}
+
+export const GenPostProvider = ({ children }: React.PropsWithChildren) => {
+  const [posts, setPosts] = useLocalStorage<GeneratedArchiveContentPost[]>(
+    'generated-archive-content-posts',
+    [],
+  )
```
7alip, line 30: It seems we don't have the control of removing posts which are generated from a certain archive content. Wouldn't be better to have the store structure like this:

```ts
type GeneratedArchiveContentPost = {
  description: string
  sentences: string[]
}

type ArchiveContentPosts = Record<number, GeneratedArchiveContentPost[]>

// Example
const storedPosts = {
  16: [
    {
      description: "Lorem Ipsum",
      sentences: ["..", "..."]
    }
  ]
}
```

With this approach, you can easily store all the generated posts for each archive content separately and be able to remove posts from a certain archive content.

7alip (commented):
Thanks

burakndrlnd (commented):
im having technical problems.

Issue #879, opened 2023-09-28
Should we consider make blog stats (likes, reads) hidden until it's fixed?


7alip:

burakndrlnd:
@7alip shold we close this ?

PR #1095, opened 2024-03-15
backend isLiked

burakndrlnd:
- added an if for development recaptcha()
- blog find() function now checks user is in likers list.and add isLiked value and save same bandwidth (likers is cutted out) and likes count is calculated here.
- like and unlike functions has changed a little bit. there is a comment.
- added user token to every blog releated strapiRequest
- and chanced icon color a little. more colorblind friendly :D i couldnt tell if i liked or not.

### DIFF
--- a/apps/api/src/api/blog/controllers/blog.ts
+++ b/apps/api/src/api/blog/controllers/blog.ts
@@ -12,5 +12,22 @@ export default factories.createCoreController('api::blog.blog', () => {
 
       return result
     },
+    async find(ctx) {
+      const result = await super.find(ctx)
+      const user = ctx.state.user
+      for (const blog of result.data) {
+        const attr = blog.attributes
+        const { likers, likes, ...others } = attr
+        const isLiked =
+          user &&
+          likers?.data?.some(liker => liker.attributes.email === user.email)
+        blog.attributes = {
+          ...others,
+          isLiked: !!isLiked,
+          likes: (likes ?? 0) + (likers?.data?.length ?? 0),
+        }
+      }
+      return result
+    },
   }
 })
--- a/apps/api/src/api/blog/controllers/blog.ts
+++ b/apps/api/src/api/blog/controllers/blog.ts
@@ -27,7 +27,7 @@ export default factories.createCoreController('api::blog.blog', () => {
 
       const isLiked =
         user &&
-        (await strapi.db.query('api::blog.blog').findOne({
+        (await strapi.db.query('api::blog.blog').count({
           where: {
             slug,
             likers: {
@@ -52,7 +52,7 @@ export default factories.createCoreController('api::blog.blog', () => {
 
         const isLiked =
           user &&
-          (await strapi.db.query('api::blog.blog').findOne({
+          (await strapi.db.query('api::blog.blog').count({
             where: {
               id,
               likers: {
--- a/apps/api/src/api/blog/controllers/blog.ts
+++ b/apps/api/src/api/blog/controllers/blog.ts
@@ -19,43 +19,59 @@ export default factories.createCoreController('api::blog.blog', () => {
 
       const blog = await strapi.db.query('api::blog.blog').findOne({
         where: { slug },
-        populate: ['author', 'likers', 'image'],
+        populate: ['author', 'image'],
       })
 
       const user = ctx.state.user
-      const { likes, likers, ...rest } = (await this.sanitizeOutput(
-        blog,
-        ctx,
-      )) as Blog
-      const isLiked = user && likers.some(liker => liker.email === user.email)
+      const sanitizedBlog = (await this.sanitizeOutput(blog, ctx)) as Blog
+
+      const isLiked =
+        user &&
+        (await strapi.db.query('api::blog.blog').findOne({
+          where: {
+            slug,
+            likers: {
+              email: user.email,
+            },
+          },
+          select: ['id'],
+        }))
 
       return {
-        ...rest,
+        ...sanitizedBlog,
         isLiked: !!isLiked,
-        likes: (likes ?? 0) + (likers?.length ?? 0),
       }
     },
     async find(ctx) {
       const response = await super.find(ctx)
       const user = ctx.state.user
 
-      response.data = response.data.map(blog => {
+      const data = []
+      for (const blog of response.data) {
         const { id, attributes } = blog
-        const { likes, likers, ...rest } = attributes
+
         const isLiked =
-          user && likers.data.some(liker => liker.email === user.email)
-        const data = {
-          ...rest,
-          isLiked: !!isLiked,
-          likes: (likes ?? 0) + (likers?.data?.length ?? 0),
-        }
+          user &&
+          (await strapi.db.query('api::blog.blog').findOne({
+            where: {
+              id,
+              likers: {
+                email: user.email,
+              },
+            },
+            select: ['id'],
+          }))
 
-        return {
+        data.push({
           id,
-          attributes: data,
-        }
-      })
+          attributes: {
+            ...attributes,
+            isLiked: !!isLiked,
+          },
+        })
+      }
 
+      response.data = data
       return response
     },
   }
--- a/apps/api/src/api/blog/controllers/blog.ts
+++ b/apps/api/src/api/blog/controllers/blog.ts
@@ -1,5 +1,6 @@
 import { factories } from '@strapi/strapi'
 import { assignCreator, getProfile } from '../../../utils'
+import { Blog } from '@fc/types'
 
 export default factories.createCoreController('api::blog.blog', () => {
   return {
@@ -12,22 +13,50 @@ export default factories.createCoreController('api::blog.blog', () => {
 
       return result
     },
-    async find(ctx) {
-      const result = await super.find(ctx)
+    async findOne(ctx) {
+      const { id } = ctx.params
+      const slug = id ? id : ctx.params.slug
+
+      const blog = await strapi.db.query('api::blog.blog').findOne({
+        where: { slug },
+        populate: ['author', 'likers', 'image'],
+      })
+
       const user = ctx.state.user
-      for (const blog of result.data) {
-        const attr = blog.attributes
-        const { likers, likes, ...others } = attr
+      const { likes, likers, ...rest } = (await this.sanitizeOutput(
+        blog,
+        ctx,
+      )) as Blog
+      const isLiked = user && likers.some(liker => liker.email === user.email)
+
+      return {
+        ...rest,
+        isLiked: !!isLiked,
+        likes: (likes ?? 0) + (likers?.length ?? 0),
+      }
+    },
+    async find(ctx) {
+      const response = await super.find(ctx)
+      const user = ctx.state.user
+
+      response.data = response.data.map(blog => {
+        const { id, attributes } = blog
+        const { likes, likers, ...rest } = attributes
         const isLiked =
-          user &&
-          likers?.data?.some(liker => liker.attributes.email === user.email)
-        blog.attributes = {
-          ...others,
+          user && likers.data.some(liker => liker.email === user.email)
+        const data = {
+          ...rest,
           isLiked: !!isLiked,
           likes: (likes ?? 0) + (likers?.data?.length ?? 0),
         }
-      }
-      return result
+
+        return {
+          id,
+          attributes: data,
+        }
+      })
+
+      return response
     },
   }
 })
--- a/apps/api/src/api/blog/controllers/custom.ts
+++ b/apps/api/src/api/blog/controllers/custom.ts
@@ -56,35 +56,22 @@ export default {
     const profile = await getProfile(ctx)
 
     if (profile) {
-      const isLikedCount = await strapi.entityService.count('api::blog.blog', {
-        filters: {
-          id: { $eq: ctx.params.id },
-          likers: { id: { $eq: profile.id } },
+      await strapi.entityService.update('api::blog.blog', ctx.params.id, {
+        data: {
+          likers: {
+            connect: [profile],
+          },
         },
       })
-
-      if (!isLikedCount) {
-        await strapi.entityService.update('api::blog.blog', ctx.params.id, {
-          data: {
-            likers: {
-              ['connect']: [profile.id as number],
-            },
-          },
-        })
-      }
+    } else {
+      // we calculate like count likes + likers.length
+      // so dont increase liker value if user added to likers
+      await strapi.db
+        .connection('blogs')
+        .where('id', ctx.params.id)
+        .increment('likes', 1)
     }
-
-    await strapi.db
-      .connection('blogs')
-      .where('id', ctx.params.id)
-      .increment('likes', 1)
-
-    const result = await strapi.entityService.findOne(
-      'api::blog.blog',
-      ctx.params.id,
-    )
-
-    return { data: result }
+    return { data: null }
   },
   async unlike(ctx: Context) {
     await checkRecaptcha(ctx)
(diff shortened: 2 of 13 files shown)


burakndrlnd:
https://github.com/freedomcombination/monorepo/issues/879

apps/api/src/api/blog/controllers/blog.ts
```diff
@@ -12,5 +12,22 @@ export default factories.createCoreController('api::blog.blog', () => {
 
       return result
     },
+    async find(ctx) {
```
7alip, line 15: On the [arts grid page](https://www.kunsthalte.com/club/arts) we allow users to be able to like without having to view the art on the detail page. In this case we should implement `isLiked` for `find` controller.

But this is not the case for blogs. Users have to visit the single blog page to be able to like/unlike the blog. It's probably best to override `findOne` for blog endpoint or create a new custom endpoint/controller for it. Here is an example: https://www.youtube.com/watch?v=OVV0CfgX6Qk

You can also consider using [sanitization and validation](https://docs.strapi.io/dev-docs/backend-customization/controllers#sanitization-and-validation-in-controllers).

apps/api/src/api/blog/controllers/custom.ts
```diff
@@ -56,71 +56,44 @@ export default {
...
-          data: {
-            likers: {
-              ['connect']: [profile.id as number],
-            },
-          },
-        })
-      }
+    } else {
+      // we calculate like count likes + likers.length
+      // so dont increase liker value if user added to likers
```
7alip, line 68: Good catch! We should then always keep in mind that `likes` count indicates the number of likes from public users.

apps/api/src/api/blog/controllers/custom.ts
```diff
@@ -56,71 +56,44 @@ export default {
...
 
     if (profile) {
-      const isLikedCount = await strapi.entityService.count('api::blog.blog', {
-        filters: {
-          id: { $eq: ctx.params.id },
-          likers: { id: { $eq: profile.id } },
+      await strapi.entityService.update('api::blog.blog', ctx.params.id, {
+        data: {
+          likers: {
+            disconnect: [profile],
```
7alip, line 85: Should it be `disconnect: [profile.id]`?

apps/api/src/api/blog/controllers/custom.ts
```diff
@@ -56,71 +56,44 @@ export default {
...
 
     if (profile) {
-      const isLikedCount = await strapi.entityService.count('api::blog.blog', {
-        filters: {
-          id: { $eq: ctx.params.id },
-          likers: { id: { $eq: profile.id } },
+      await strapi.entityService.update('api::blog.blog', ctx.params.id, {
+        data: {
+          likers: {
+            connect: [profile],
```
7alip, line 62: How it should work? Maybe `connect: [profile.id]`?

apps/foundation/pages/blog/[slug].tsx
```diff
@@ -51,17 +53,24 @@ export const getServerSideProps = async (
   const slug = context.params?.['slug'] as string
   const queryKey = ['blog', locale, slug]
 
+  const { token } = await getSession(context.req, context.res)
+
   await queryClient.prefetchQuery({
     queryKey,
-    queryFn: () => getBlogBySlug(locale, slug),
+    queryFn: () => getBlogBySlug(locale, slug, token),
```
Comment (7alip) [inline apps/foundation/pages/blog/[slug].tsx:60]: Shouldn't we keep some endpoints as public. Can public users still view the blog page?

apps/foundation/pages/blog/[slug].tsx
```diff
@@ -51,17 +53,24 @@ export const getServerSideProps = async (
...
 
   if (!blog) return { notFound: true }
 
   const authorBlogs =
-    (await getAuthorBlogs(locale, blog?.author?.id as number, blog.id)) || []
+    (await getAuthorBlogs(
+      locale,
+      blog?.author?.id as number,
+      blog.id,
+      token,
```
Comment (7alip) [inline apps/foundation/pages/blog/[slug].tsx:72]: The same. Shouldn't we make this public?

apps/foundation/pages/blog/index.tsx
```diff
@@ -26,14 +29,17 @@ const Blogs = () => {
...
+) => {
   const queryClient = new QueryClient()
 
   const locale = context.locale as StrapiLocale
+  const { token } = await getSession(context.req, context.res)
 
   await queryClient.prefetchQuery({
     queryKey: ['blogs', locale],
-    queryFn: () => getBlogs(locale),
+    queryFn: () => getBlogs(locale, token),
```
7alip, line 42: The same public endpoint

packages/services/src/blog/get.ts
```diff
@@ -1,14 +1,16 @@
...
 import { strapiRequest } from '@fc/lib'
 import { Blog, StrapiLocale } from '@fc/types'
 
-export const getBlogs = async (locale: StrapiLocale) => {
+export const getBlogs = async (locale: StrapiLocale, token: string | null) => {
   const response = await strapiRequest<Blog>({
     endpoint: 'blogs',
     locale,
     sort: ['publishedAt:desc'],
+    ...(token && { token }),
```
7alip, line 13: Blogs can remain as public

7alip (commented):
Thanks

apps/api/src/api/blog/controllers/blog.ts
burakndrlnd: in the begining, i override findOne but none of blog request has triggered to findOne(). then i switch to find(). i mean before i change something frontend used that endpoint.

apps/api/src/api/blog/controllers/custom.ts
burakndrlnd: after overriding of fine(), 'likes' now indicates total like count.
` likes: (likes ?? 0) + (likers?.data?.length ?? 0), `
this code used in BlogEdit. i removed whole likers from endpoint response. only data we get from strapi total like count in 'likes' variable, and 'isLiked' if likers contains user.

apps/api/src/api/blog/controllers/custom.ts
burakndrlnd: https://docs.strapi.io/dev-docs/api/rest/relations#connect
shorthand [ number ] , longhand [ { id: number } ] 
both of them are okey.

apps/api/src/api/blog/controllers/blog.ts
7alip: Because currently your endpoint is `/api/blogs` not `/api/blogs/:slug`.

Current `getBlogBySlug` function is fetching from `/api/blogs` (find) endpoint.

```ts
export const getBlogBySlug = async (
  locale: StrapiLocale,
  slug: string,
): Promise<Blog> => {
  const response = await strapiRequest<
    SetRequired<Blog, 'author' | 'image' | 'likers'>
  >({
    endpoint: 'blogs',
    populate: ['author', 'image', 'likers'],
    filters: { slug: { $eq: slug } },
    locale,
  })

  return response?.data?.[0] || null
}
```

You can maybe do some hack like below to trigger `/api/blogs/:slug` (findOne).

```ts
export const getBlogBySlug = async (
  slug: string,
): Promise<Blog> => {
  const response = await strapiRequest<
    SetRequired<Blog, 'author' | 'image' | 'likers'>
  >({
    endpoint: 'blogs',
    id: slug as unknown as number,
    populate: ['author', 'image', 'likers'],
  })

  return response?.data || null
}
```

apps/foundation/pages/blog/[slug].tsx
Comment (burakndrlnd) [reply apps/foundation/pages/blog/[slug].tsx:60]:
```
const result = await super.find(ctx)
const user = ctx.state.user
for (const blog of result.data) {
  const attr = blog.attributes
  const { likers, likes, ...others } = attr
  const isLiked =
    user &&
    likers?.data?.some(liker => liker.attributes.email === user.email)
  blog.attributes = {
    ...others,
    isLiked: !!isLiked,
    likes: (likes ?? 0) + (likers?.data?.length ?? 0),
  }
}
return result
```
if `token` is null `ctx.state.user` becomes null. then isLiked ll be false. so our endpoints stay as before.

apps/api/src/api/blog/controllers/custom.ts
7alip: Than it's probably safer to use shorthand since the profile has other fields

`disconnect: [{ id: 1, username: "profile", ... }]`

apps/foundation/pages/blog/[slug].tsx
Comment (7alip) [reply apps/foundation/pages/blog/[slug].tsx:60]: Ah, of course. I totally forgot about isLiked field :)

Comment (7alip) [reply apps/foundation/pages/blog/[slug].tsx:60]: @burakndrlnd We probably need to pass token only to `getBlogBySlug` not the other blog fetchers.

apps/foundation/pages/blog/index.tsx
burakndrlnd: image

as you can see, not logged in and still viewable like before. if i get it right from 'public'.

burakndrlnd:
i had to override both `find()` and `findOne()`. 
`findOne()` is modified to do thing the way that you ask to do.
and
`find()` is modified because in frontend `/blog/index.tsx` there is `BlogCard` and it shows like count as well. but it never regarded `likers.count`.  now likers.count added to likes in strapi.

apps/api/src/api/blog/controllers/custom.ts
```diff
@@ -56,71 +56,44 @@ export default {
...
-      const isLikedCount = await strapi.entityService.count('api::blog.blog', {
-        filters: {
-          id: { $eq: ctx.params.id },
-          likers: { id: { $eq: profile.id } },
+      await strapi.entityService.update('api::blog.blog', ctx.params.id, {
+        data: {
+          likers: {
+            disconnect: [profile.id],
+          },
         },
```
7alip, line 87: If `likes` indicates total likes from both authenticated and public users, shouldn't we decrease `likes` count in this if statement as well?

apps/api/src/api/blog/controllers/custom.ts
```diff
@@ -56,71 +56,44 @@ export default {
...
     if (profile) {
-      const isLikedCount = await strapi.entityService.count('api::blog.blog', {
-        filters: {
-          id: { $eq: ctx.params.id },
-          likers: { id: { $eq: profile.id } },
+      await strapi.entityService.update('api::blog.blog', ctx.params.id, {
+        data: {
+          likers: {
+            connect: [profile.id],
+          },
```
7alip, line 63: If likes indicates total likes from both authenticated and public users, shouldn't we increase likes count in this if statement as well?

If we always update `likes` filed at this point, we don't have to manipulate the response.

apps/api/src/api/blog/controllers/blog.ts
```diff
@@ -12,5 +13,50 @@ export default factories.createCoreController('api::blog.blog', () => {
...
+    },
+    async find(ctx) {
+      const response = await super.find(ctx)
+      const user = ctx.state.user
+
+      response.data = response.data.map(blog => {
+        const { id, attributes } = blog
+        const { likes, likers, ...rest } = attributes
+        const isLiked =
+          user && likers.data.some(liker => liker.email === user.email)
```
7alip, line 46: I'm not sure what's the best approach for checking isLiked. For example, can the following approach be more performant?

```ts
const likedBlogIdsByUser = await strapi.entityService.findMany('api:blog.blog', {
  filters: { likers: { email: { $eq: user.email } } },
  fields: ['id']
})

return {
  ...response,
  data: response.data.map(blog => {
    const { id, attributes } = blog
    const isLiked = likedBlogIds.includes(blog.id)
    
    // If we increase or decrease the `likes` field on like/unlike action, 
    // we don't need to sum `likes + likers.length` here
    // Otherwise, we have to always populate `likers` field.
    return { ...blog, isLiked }
  })
}
```

7alip (commented):
Thanks, I left some comments. I'm curious what do you think about the approaches.

apps/api/src/api/blog/controllers/custom.ts
burakndrlnd: no. because its only calculated in `find()/findOne()` functions.

in original code (i mean main branch) `likes` variable represent public users and `likers.length` represent registered user and they are still same.

in `BlogDetail.tsx` , this is how like count calculated:
image

i just move that calculation to backend : 
image

i seems to do that is good idea.

apps/api/src/api/blog/controllers/custom.ts
burakndrlnd: as i mention above. this is a choose. i tried to change things as little as possible from original. that's way i stick this way.

apps/api/src/api/blog/controllers/blog.ts
burakndrlnd: if we have 1000 likes per blog, yes this would be better way by far. but in our stuation i believe the fewer queries the better.

and making benchmark with strapi backend is pain. not even benchmark, even writing is. appling even little changes takes a minute. i dont know if im the problem.

by the way, I accidentally opened the pull request as a draft and I couldnt change it, I dont know if it makes any difference, so im noting it here.

apps/api/src/api/blog/controllers/custom.ts
7alip: When you begin reviewing the code, you will notice your own mistakes as well. Or you can think differently at some point. That has happened to me here.

When I first implemented this I probably thought that `likes` should indicate the public users but I think differently now :D

Updating the `likes` count always sounds good too, and getting the `likes` count without having to populate `likers` seems like a better idea to me now.

apps/api/src/api/blog/controllers/blog.ts
7alip: We probably won't have many authenticated users, so your implementation is perfectly fine. I was just thinking about the proper way to handle this type of situation.

___

Regarding the slowness, it appears to be memory-related, and I believe our project may be experiencing some ram usage issues. I probably shouldn't have exported the files from `index.ts` files in every folder throughout the project.

___

Normally you can remove draft mode with `Ready for review` button but I'm not sure if it's visible for you.

Screenshot 2024-03-17 at 04 32 33

7alip:
@burakndrlnd There is linting issue in the pipeline in `pages/blog/[slug].tsx`

Screenshot 2024-03-17 at 04 55 49

apps/api/src/api/blog/controllers/custom.ts
burakndrlnd: > When you begin reviewing the code, you will notice your own mistakes as well.
happens all the time :)

i ll make the changes.

apps/api/src/api/blog/controllers/blog.ts
burakndrlnd: > Regarding the slowness

i take this screenshot if it means anything to you : 
image
and right bottom, that lint was working fine 2 days ago.

> you can remove draft mode with Ready for review button

yes it is visible but i didnt know its usage thanks.

7alip: 62 sec is insane! We should definitely investigate about this issue. I'm using mac with 32GB ram. Maybe that's why I'm not facing with this issue.

___

<details>
  <summary>Lint error</summary>
Screenshot 2024-03-17 at 12 05 41

</details>

apps/api/src/api/blog/controllers/blog.ts
```diff
@@ -12,5 +13,66 @@ export default factories.createCoreController('api::blog.blog', () => {
...
+      const response = await super.find(ctx)
+      const user = ctx.state.user
+
+      const data = []
+      for (const blog of response.data) {
+        const { id, attributes } = blog
+
+        const isLiked =
+          user &&
+          (await strapi.db.query('api::blog.blog').findOne({
```
7alip, line 55: I don't think it's necessary to expand blogs response (find), but we can show the heart icon as red on the blog cards. Let's keep it then.

`.count` could be more performant.

7alip (approved):
Thanks

7alip (approved):
Thanks

Issue #945, opened 2023-11-04
Fix used email error notification in voluntary form


7alip:
<img width="1431" alt="Screenshot 2023-11-04 at 14 19 13" src="https://github.com/wsvvrijheid/monorepo/assets/22167684/e5cb558b-9f1c-4c6e-a177-da362fb20b81">

Fix translation
Preferably trigger email form error which will appear underneath email input

PR #1097, opened 2024-03-15
application form translate

burakndrlnd:
https://github.com/freedomcombination/monorepo/issues/945

by the way, i couldnt reproduce same warning at issue. but translation is okey now.


7alip (approved):
Thanks

PR #1107, opened 2024-03-20
art-like .

burakndrlnd:


burakndrlnd:
image

by the way i added 32gb ram today and now its better. i think the main reason was garbace collector. 40 seconds is from scratch. if i edit some strapi file then it takes 5-6 secs.

7alip (commented):
I created another PR to fix cache issues.

I also noticed that when I like an art from the list (not the detail page), like counts are not changing. Maybe you can check this as well.

PR #1124, opened 2024-03-27
platform coordinator role added

MustafaTRKYLMZ:
Platform coordinator role can see:

- [ ] profiles mostly volunteers, If it is possible, they can see according to filter platform. For example, lotus vd media coordinator can see only who applied as volunteer for lotus vd media. 
- [x] courses CRU, which is releated with platform.


7alip (changes requested):
If `platformcoordinator` should only see volunteers (profiles) associated with the coordinator's platform, more controller-level adjustments are required. You should adjust the profile `find` controller with a special filter for this role. This part can be a bit complex. There are some examples in the hashtag-reports project where we added extra filters for location admins.

___  

There is also TS issues in the pipeline

Screenshot 2024-03-27 at 03 29 37

7alip:
@MustafaTRKYLMZ 

Screenshot 2024-03-31 at 10 50 35

MustafaTRKYLMZ:
Sorry, for that. There is a bug with my account. That is why, I couldn't test. 
image

7alip:
> Sorry, for that. There is a bug with my account. That is why, I couldn't test.

@MustafaTRKYLMZ did you check from `localhost:1337` if your account exists and has the correct role?

7alip (approved):
Thanks

PR #1128, opened 2024-03-27
Login redirect

7alip:


packages/ui/src/components/CreateArtForm/CreateArtForm.tsx
```diff
@@ -80,6 +81,7 @@ export const CreateArtForm = () => {
   const isValid = formState.isValid
 
   const { mutate, isPending } = useCreateModelMutation('arts')
+  const loginHref = getLoginUrl(useRouter())
```
7alip, line 84: You could use a custom hook `useLoginUrl` and useRouter could be imported in it.

You can also place the hook in `packages/ui/src/hooks`

packages/ui/src/components/LoginForm/getLoginUrl.ts
```diff
@@ -0,0 +1,3 @@
+
+export const getLoginUrl = ({ pathname }: { pathname: string }) =>
+    `/login?returnUrl=${encodeURIComponent(pathname)}`
```
7alip, line 3: Isn't `pathname` already encoded? Do we need to encode it?

```tsx
export const useLoginUrl = () => {
  const { pathname } = useRouter()
  
  return `/login?returnUrl=${encodeURIComponent(pathname)}`
}
```

PR #1182, opened 2024-04-10
Refactor PM managing sentences

7alip:


packages/ui/src/post-maker/PostMakerTweetCard/PostMakerTweetShare.tsx
```diff
@@ -18,21 +18,14 @@ import {
...
 export const PostMakerTweetShare: FC<PostMakerTweetShareProps> = ({
   url,
   content,
+  allowManagingSentences,
 }) => {
-  const { roles, checkAuth } = useAuthContext()
-  useEffect(() => {
-    checkAuth()
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [])
```
7alip, line 34: This part seems to be overkill

myniqx (approved):
to test it,
i create a new hashtag then approved / published it.
then i create 7 post and in it 7 sentences, each approved / published both posts and sentences in them.
but in post maker, i only see hashtag. when i click see more there is no post/sentences in it. 
other hashtags are fine. but i want to test it with alot of sentences on local, this time and for future. which step did i miss?

7alip:
> to test it, i create a new hashtag then approved / published it. then i create 7 post and in it 7 sentences, each approved / published both posts and sentences in them. but in post maker, i only see hashtag. when i click see more there is no post/sentences in it. other hashtags are fine. but i want to test it with alot of sentences on local, this time and for future. which step did i miss?

I'm not sure why the posts are not visible. You need to debug it and see what do the requests return in the getServerSideProps of the hashtag page: `apps/trend-rights/pages/hashtags/[slug]/index.tsx`

Issue #951, opened 2023-11-06
Make tweets and news menus visible for authenticated users only


MustafaTRKYLMZ:
- [ ] Make it visible in the header
- [ ] Make it visible in the footer

PR #1197, opened 2024-04-14
Burak-trend-menus

myniqx:
https://github.com/freedomcombination/monorepo/issues/951

### DIFF
--- a/apps/trend-rights/components/Layout/Layout.tsx
+++ b/apps/trend-rights/components/Layout/Layout.tsx
@@ -3,6 +3,7 @@ import { FC, PropsWithChildren, useEffect } from 'react'
 import { NextSeoProps } from 'next-seo'
 
 import { menus, socialLinks } from '@fc/config'
+import { news, tweets } from '@fc/config/src/menu/routes'
 import { useAuthContext } from '@fc/context'
 import { Layout as AppLayout, UserFeedback, useScroll } from '@fc/ui'
 
@@ -18,7 +19,7 @@ export const Layout: FC<LayoutProps> = ({
   isLoading,
   seo,
 }) => {
-  const { checkAuth } = useAuthContext()
+  const { checkAuth, token: isLoggedIn } = useAuthContext()
   const isScrolled = useScroll()
 
   useEffect(() => {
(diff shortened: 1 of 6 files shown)


apps/trend-rights/components/Layout/Layout.tsx
```diff
@@ -18,20 +19,28 @@ export const Layout: FC<LayoutProps> = ({
...
-  const { checkAuth } = useAuthContext()
+  const { checkAuth, token: isLoggedIn } = useAuthContext()
   const isScrolled = useScroll()
 
   useEffect(() => {
     checkAuth()
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [])
 
+  let headerMenu = menus['trend-rights'].headerMenu
```
7alip, line 30: Here is what GPT says about this approach:

<blockquote>In a React component, if the change in headerMenu and footerMenu based on isLoggedIn should cause a re-render of the component, then you should use useState. This is because changes in state variables in React cause the component to re-render, and this is the recommended way to handle changes that should reflect in the UI.

Here's how you could do it:

```tsx
import React, { useState, useEffect } from 'react';

// ... other code ...

const [headerMenu, setHeaderMenu] = useState(menus['trend-rights'].headerMenu);
const [footerMenu, setFooterMenu] = useState(menus['trend-rights'].footerMenu);

useEffect(() => {
  if (isLoggedIn) {
    setHeaderMenu([news, tweets, ...menus['trend-rights'].headerMenu]);
    setFooterMenu(prevFooterMenu => {
      const newFooterMenu = [...prevFooterMenu];
      newFooterMenu[0].children = [news, tweets, ...(newFooterMenu[0].children ?? [])];
      
      return newFooterMenu;
    });
  }
}, [isLoggedIn]); // re-run this effect when isLoggedIn changes
```

This way, whenever isLoggedIn changes, the effect will run, updating headerMenu and footerMenu and causing a re-render if their values have changed.
</blockquote>

7alip (commented):
Thanks

myniqx:
i tried to keep the menus a little more tidy. it can be modified from a single place (packages\config\src\menu\trend-rights.ts).

7alip (approved):
Thanks

Issue #1155, opened 2024-04-07
Display an active hashtag modal when users first visit the TrendRights website during the hashtag event.


7alip:
## Description

Enhance website functionality to automatically redirect users to the PostMaker page during active hashtag events. Implement this feature universally across all pages using Next.js middleware, leveraging Vercel's `Edge Config` and cookies for efficient execution.

> I created a edge config called `last-hashtag` from Vercel dashboard. 
Config Id: `ecfg_teewtrjl8mpox0ulyt68bd43hgbi`. 
Some new env variables needs to be added to env.local 

## Tasks

- [ ] **[Update](https://vercel.com/docs/storage/edge-config/vercel-api#update-your-edge-config-items)** `last-hashtag` edge config key whenever an hashtag is created.
  _This can be managed through Strapi's lifecycle hooks (afterCreate, afterUpdate)_
  _The config value should contain the hashtag's slug and hashtag's date in ISO format (e.g., `patient-rights__2024-04-07T10:19:00.000Z`)._
- [ ] Use two cookies
  1. `pm-ir` (boolean) (_post maker is redirected_)
  2. `pm-lhs` (_post maker lasthashtag slug)_)
  _Set this `pm-ir` cookie to true when user closes the modal or clicks on the PM redirect link and update `pm-lhs` cookie with the edge config slug_
- [ ] In Next.js middleware, check if the current ISO time is within the latest active hashtag event and compare cookie-config slugs. 
  _Hashtag datetime + 12 hours_
- [ ] Trigger the hashtag modal conditionally based on `pm-ir` cookie in the layout 

## Example

```ts
import { NextResponse } from 'next/server';
import { get } from '@vercel/edge-config';

export async function middleware() {
  const lastHashtag = await get('last-hashtag');
  
  const [slug, dateStr] = (lastHashtag || '').split('__');
  // ...
}
```

PR #1198, opened 2024-04-14
Burak-last-hashtag-redirect

myniqx:
https://github.com/freedomcombination/monorepo/issues/1155

### DIFF
--- a/apps/trend-rights/middleware.ts
+++ b/apps/trend-rights/middleware.ts
@@ -1,3 +1,4 @@
+import { get } from '@vercel/edge-config'
 import { addHours, isWithinInterval, parseISO } from 'date-fns'
 import { NextRequest, NextResponse } from 'next/server'
 
@@ -7,11 +8,9 @@ const corsHeaders = {
   'Access-Control-Allow-Headers': 'Content-Type, Authorization',
 }
 
-// this cookie ll hold last redirected slug's name
-const cookieRedirectedSlug = '1redirected-slug'
-
-// this cookie ll have a minute long life. if this cookie exists, it means that we already fetched data
-const cookieDoFetch = '1do-fetch'
+const SLUG_COOKIE_KEY = 'hashtag-redirect-slug'
+const EDGE_CONFIG_KEY =
+  process.env.NODE_ENV === 'production' ? 'last-hashtag' : 'last-hashtag-dev'
 
 export async function middleware(request: NextRequest) {
   if (request.nextUrl.pathname.startsWith('/api/')) {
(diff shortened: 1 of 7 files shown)


apps/trend-rights/middleware.ts
```diff
@@ -6,18 +8,42 @@ const corsHeaders = {
   'Access-Control-Allow-Headers': 'Content-Type, Authorization',
 }
 
-export function middleware(request: NextRequest) {
-  if (request.method === 'OPTIONS') {
-    return NextResponse.json({}, { headers: corsHeaders })
+const dontRedirectIfDevelopment = true // prevents redirecting if in development
```
7alip, line 11: Let's redirect in dev mode as well. We can easily check it always

apps/trend-rights/middleware.ts
```diff
@@ -6,18 +8,42 @@ const corsHeaders = {
...
+  const hashtagDatetimePlus12Hours = addHours(hashtagDatetime, 12)
+
+  const currentDate = new Date()
+  const isCurrentDateWithinInterval = isWithinInterval(currentDate, {
+    start: hashtagDatetime,
+    end: hashtagDatetimePlus12Hours,
+  })
+
+  if (isCurrentDateWithinInterval) {
+    return NextResponse.redirect('/hashtags/' + slug)
```
7alip, line 47: https://nextjs.org/docs/pages/building-your-application/routing/redirecting#nextresponseredirect-in-middleware

In Next.js pages router example, they use redirect like this:

`return NextResponse.redirect(new URL('/login', request.url))`

apps/trend-rights/middleware.ts
```diff
@@ -6,18 +8,42 @@ const corsHeaders = {
...
+  const hashtagDatetime = parseISO(dateStr)
+  const hashtagDatetimePlus12Hours = addHours(hashtagDatetime, 12)
+
+  const currentDate = new Date()
+  const isCurrentDateWithinInterval = isWithinInterval(currentDate, {
+    start: hashtagDatetime,
+    end: hashtagDatetimePlus12Hours,
+  })
+
+  if (isCurrentDateWithinInterval) {
```
7alip, line 46: Check if redirected already

```ts
const isRedirected = slug === cookieSlug
```

7alip (commented):
Thanks

myniqx:
@7alip im sorry i didnt get review request back. i rewrote almost everthing. its better.
what is working:
- reads from edge-config 
- if it tries to fetch the key more than once in 10 seconds, it skips (cookie named dofetch)
- if it already redirected (cookie named redirectedslug) or already in hashtags/:slug, it skips.
- in backend afterCreate and afterUpdate without check any condition last hashtag ,in teory, added to edge config.
what is not working:
-last hashtag cannot be updated to edge config.

i test it under there condition:
- i return static last-hashtag and do testing and get those result (above)
- and also backend works until set function. it returns an error 403. it seems that token in connection string is working while reading, but when create or update, it doesnt work.
if you paste the connection string you gave me to address bar you can easly see its working.
but when write (https://vercel.com/docs/storage/edge-config/vercel-api#update-your-edge-config-items)
this is the exact same error : 
```
{
  "error": {
    "code": "forbidden",
    "message": "The request is missing an authentication token",
    "missingToken": true
  }
}
```
maybe with the correct token, this ll work.

myniqx:
@7alip now its better :)
in middleware:
i added matcher, with a litlle bir more filter it works without do-fetch cookie. so i remove it.
vercel's get function is now working. it seems it was a matter releated with turbo. 
and i added another edge key for development mode, 'last-hashtag-dev'.

7alip (approved):
Thanks

PR #1211, opened 2024-04-16
Burak-pm-sentence

myniqx:
it was contenteditable after all.
at the begining it sets his currentValue with empty string (i think its happening because sentences hasnt finished fetching)
then comes real value. but then 700ms later debounce occurs and calls onupdate so it sets that sentence null. now its fixed. but only for this case.

and the changes in getBySlug, they are releated to posts that generated by PostGenAI. we set their imageparams but getBySlug filters them by image or caps. now they are ready to see after they published. (it uses hashtag's image by default)

### DIFF
--- a/packages/services/src/hashtag/getBySlug.ts
+++ b/packages/services/src/hashtag/getBySlug.ts
@@ -34,17 +34,18 @@ export const getHashtagBySlug = async (
 
   const posts =
     hashtag.posts
-      ?.filter(p => p.image || p.caps)
+      ?.filter(p => p.image || p.caps || p.imageParams)
       .map((p, index) => ({
         ...p,
         index,
-        image: {
+        image: p.image || p.caps ? {
           url:
             p.image?.formats?.small?.url ||
             p.image?.formats?.medium?.url ||
             p.image?.url ||
             p.caps?.url,
-        } as UploadFile,
+        } as UploadFile :
+          p.imageParams as UploadFile,
       })) || []
 
   const localizations = (hashtag.localizations?.map(l => ({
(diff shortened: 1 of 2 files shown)


packages/services/src/hashtag/getBySlug.ts
```diff
@@ -34,17 +34,18 @@ export const getHashtagBySlug = async (
...
-        image: {
+        image: p.image || p.caps ? {
           url:
             p.image?.formats?.small?.url ||
             p.image?.formats?.medium?.url ||
             p.image?.url ||
             p.caps?.url,
-        } as UploadFile,
+        } as UploadFile :
+          p.imageParams as UploadFile,
```
7alip, line 48: We expect the post.image as { url: string } format, if we use imageParams here, how it's being used now?

myniqx: there are 2 points here;
first; logic hasnt changed. if there is no image or caps then it goes for imageParams (if imageParam exist). earlier if there is no image or caps then it was filtered... so there are no effect on old entities.
second: in imageParams:
```
{
  "bg": "#FAF5FF",
  "flip": false,
  "color": "#805AD5",
  "image": {
    "id": 782,
    "ext": ".jpeg",
    "url": "/uploads/blob_d41ecf506d.jpeg",
    "hash": "blob_d41ecf506d",
    "mime": "image/jpeg",
    "name": "blob",
    "size": 436.65,
``` 
{ url : string } already exist.

and result is;
this is hashtags page (hashtag with its image): 
image
this is hashtags posts page( post with default(hashtags image) and their own images)
image
first and third have no own image so they use imageParam (hashtags image) second has own image

7alip (approved):
Thanks

Issue #1199, opened 2024-04-14
Clear session cookie on logout from TrendRights website


MustafaTRKYLMZ:
We noticed that when you log out of the TrendRights website, the iron-session cookie doesn't seem to be removed.

PR #1212, opened 2024-04-16
trend-right logout

myniqx:
https://github.com/freedomcombination/monorepo/issues/1199

and added loading spinner to login button


7alip (approved):
Thanks

PR #1214, opened 2024-04-17
Fix edge config

7alip:
Could you look into this? I left some comments in the code

### DIFF
--- a/apps/api/src/api/hashtag/content-types/hashtag/lifecycles.ts
+++ b/apps/api/src/api/hashtag/content-types/hashtag/lifecycles.ts
@@ -1,8 +1,14 @@
 const EDGE_CONFIG_KEY =
-  process.env.NODE_ENV === 'production' ? 'last-hashtag' : 'last-hashtag-dev'
-const set = async (value: string) => {
+  process.env.VERCEL_ENV === 'production' ? 'last-hashtag' : 'last-hashtag-dev'
+
+// TODO: Update the edge config based on hashtag's locale
+// We need to first fetch the config items, then update the one we need
+// And what if the updated/created hashtag is newer than one of the upcoming events?
+// We should probably update the edge config only if the hashtag is closer to the current date
+// In this case, how do we handle updating for the latest hashtags which have created/updated in the past?
+const updateEdgeConfig = async (value: string) => {
   try {
-    const requestUrl = `https://api.vercel.com/v1/edge-config/${process.env.EDGE_TOKEN}/items`
+    const requestUrl = process.env.EDGE_ENDPOINT
     const requestConfig = {
       method: 'PATCH',
       headers: {
@@ -36,11 +42,7 @@ const set = async (value: string) => {
       )
     }
 
-    console.info(
-      "Updated edge config '",
-      EDGE_CONFIG_KEY,
-      "' to '" + value + "'",
-    )
+    console.info(`Updated edge config '${EDGE_CONFIG_KEY}' to '${value}'`)
 
     return result
   } catch (error) {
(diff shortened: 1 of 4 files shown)


apps/api/src/api/hashtag/content-types/hashtag/lifecycles.ts
```diff
@@ -1,8 +1,14 @@
 const EDGE_CONFIG_KEY =
-  process.env.NODE_ENV === 'production' ? 'last-hashtag' : 'last-hashtag-dev'
-const set = async (value: string) => {
+  process.env.VERCEL_ENV === 'production' ? 'last-hashtag' : 'last-hashtag-dev'
+
+// TODO: Update the edge config based on hashtag's locale
+// We need to first fetch the config items, then update the one we need
+// And what if the updated/created hashtag is newer than one of the upcoming events?
+// We should probably update the edge config only if the hashtag is closer to the current date
+// In this case, how do we handle updating for the latest hashtags which have created/updated in the past?
```
myniqx, line 8: This came to my mind too. since it was not mentioned in the issue, i thought that there is no such confusion when using it. to solve this, hashtag query can be made with date filter in afterCreate/Update and the result of that query can be passed as a parameter.

apps/api/src/api/hashtag/content-types/hashtag/lifecycles.ts
```diff
@@ -50,12 +52,12 @@ const set = async (value: string) => {
...
-    const edgeValue = `${result.slug}__${result.date}`
-    await set(edgeValue)
+    const edgeValue = `${result.id}__${result.date}`
+    await updateEdgeConfig(edgeValue)
   },
 
   async afterUpdate({ result }) {
-    const edgeValue = `${result.slug}__${result.date}`
-    await set(edgeValue)
+    const edgeValue = `${result.id}__${result.date}`
```
myniqx, line 60: if we use id here, dont we need to update middleware too?

apps/trend-rights/middleware.ts
```diff
@@ -27,35 +28,32 @@ export async function middleware(request: NextRequest) {
 
   const lastHashtag = await get(EDGE_CONFIG_KEY)
   const [slug, dateStr] = ((lastHashtag as string) || '').split('__')
+  const hashtagSlug = `/hashtags/${slug}`
```
myniqx, line 31: is 'hashtags/65' valid route?

apps/trend-rights/middleware.ts
```diff
@@ -27,35 +28,32 @@ export async function middleware(request: NextRequest) {
...
-      new URL('/hashtags/' + slug, request.url),
-    ).cookies.set({
-      name: SLUG_COOKIE_KEY,
-      value: slug,
-      maxAge: 12 * 60 * 60,
-    })
+    // FIXME: Redirecting and cookie setting didn't work for me
+    response.cookies.set(SLUG_COOKIE_KEY, slug)
+
+    return NextResponse.redirect(new URL(hashtagSlug, request.url))
```
myniqx, line 53: In this case, we set cookie to response which we wont send. But gpt says its okey.

apps/api/src/api/hashtag/content-types/hashtag/lifecycles.ts
7alip: @myniqx Sorry, I forgot to revert this, first I thought that `id` field is the same for each localized entries, but it's not the case. It should be slug.

apps/trend-rights/middleware.ts
7alip: As I mentioned, I tried `getHashtagBySlugOrId` but then I realised that `id` is also different for each language.

apps/trend-rights/middleware.ts
7alip: What you say can be correct. I changed it because this was giving error in the console: 
```ts
return NextResponse.redirect(new URL(hashtagSlug, request.url)).cookies.set(SLUG_COOKIE_KEY, slug)
```

apps/api/src/api/hashtag/content-types/hashtag/lifecycles.ts
7alip: Let's keep it simple. Most of time we update the `hashtag` field of the hashtag just before the hashtag event is started. Perhaps this will never be an issue, we can easily fix it by updating the hashtag on time.

Issue #737, opened 2023-07-25
Improve Caps text design


7alip:

myniqx:
can i have some details. which way do you want it to be improved? centered text or dynamic font size ?

7alip:
Centered text would be fine.

We should consider how it looks like with long descriptions (200-240 chars). You can add some stories with variations.

og api route should also be adjusted in the same way

PR #1219, opened 2024-04-17
Burak-caps-text

myniqx:
https://github.com/freedomcombination/monorepo/issues/737

image

image

image

image

### DIFF
--- a/packages/services/src/api/og.tsx
+++ b/packages/services/src/api/og.tsx
@@ -230,6 +230,9 @@ export const ogRouter = async (req: NextRequest) => {
                 fontSize: `${dimensions.font}px`,
                 overflow: 'hidden',
                 whiteSpace: 'pre-wrap',
+                textAlign: 'center',
+                flexGrow: 1,
+                alignItems: 'center',
               }}
             >
               {text}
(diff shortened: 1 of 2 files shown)


7alip (changes requested):
Thanks.

The Caps component is displayed on the dashboard and TrendRights cards. However, on Twitter, users will see the SEO OpenGraph image, which is more important.

This style improvement should also be applied to `packages/services/src/api/og.tsx`.

To test the og image path, use the '' component with the ogParams query.

myniqx:
i updated to releated changes, is there any way to check is it working correctly ?

packages/services/src/api/og.tsx
```diff
@@ -230,6 +230,9 @@ export const ogRouter = async (req: NextRequest) => {
                 fontSize: `${dimensions.font}px`,
                 overflow: 'hidden',
                 whiteSpace: 'pre-wrap',
+                textAlign: 'center',
```
7alip, line 233: @myniqx Shall we make it not center but left aligned? It looks not good if the last line has few words.

7alip (approved):
Thanks

packages/services/src/api/og.tsx
myniqx: what about this ?
`textAlign: flip ? 'right' : 'left',`

7alip (approved):
Thanks

packages/services/src/api/og.tsx
7alip: Let's keep it like this. We can update this logic when we start implementing new templates

PR #1234, opened 2024-04-22
Burak-platform-logo

myniqx:
i have tested with caps. but i dont know how to run og.tsx

https://github.com/freedomcombination/monorepo/issues/1163


packages/ui/src/components/Caps/Caps.tsx
```diff
@@ -33,6 +34,7 @@ export const Caps = forwardRef<CapsProps, 'div'>(
       color,
       flip = false,
       hasLine = false,
+      platform = 'trend-rights',
```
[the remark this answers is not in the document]
7alip: I know it was mentioned in the task: `Add platform: PlatformSlug to OgImageParams (default is trend-rights)` but I noticed that if `platform` will be part of `imageParams`, it will be difficult to update it when we change hashtag's platform.

It's probably best to get platform info directly from `post.hashtag.platform.slug`. Even if we don't save platform in `post.imageParams`, we can still extend imageParams with platform in `getHashtagBySlug`.

packages/ui/src/components/Caps/logos.tsx
```diff
@@ -0,0 +1,1022 @@
...
+            width="452"
+            height="452"
+            fill="white"
+            transform="translate(30 30)"
+          />
+        </clipPath>
+      </defs>
+    </>
+  ),
+  'trend-rights': (
```
7alip, line 47: I see it's previous filled logo

7alip (changes requested):
Thanks

packages/ui/src/components/Caps/logos.tsx
myniqx: i took lines from file. i didnt know they are different.

packages/services/src/api/og.tsx
```diff
@@ -22,6 +28,10 @@ export const ogRouter = async (req: NextRequest) => {
   const flip = params.get('flip') ? params.get('flip') === 'true' : false
   const hasLine =
     title && params.get('hasLine') ? params.get('hasLine') === 'true' : false
+  const platformSlug = (params.get('platform') ??
+    'trend-rights') as PlatformSlug
```
7alip, line 32: @myniqx How we hide logo in og.tsx?

myniqx: I thought its enough to hide it in <Caps/> on the site. fixing it right away.

7alip (approved):
Thanks

Issue #1226, opened 2024-04-20
Filter profiles endpoint for PlatformCoordinator role


7alip:
### Is your feature request related to a problem? Please describe.

The PlatformCoordinator role can view profiles on the dashboard. They can now see all profiles, but they should only be able to see profiles from their own platforms (coordinator.platform).

### Describe the solution you'd like

Override `find` controller of the profile endpoint. `apps/api/src/api/profile/controllers/profile.ts`. All other filters, population, and sorting should work as usual.

Use sanitize query and extend the filters.

`const sanitizedQueryParams = await this.sanitizeQuery(ctx)`

### Additional context

https://docs.strapi.io/dev-docs/backend-customization/controllers

PR #1235, opened 2024-04-23
filtered profiles endpoint for PlatformCoordinator role

myniqx:
https://github.com/freedomcombination/monorepo/issues/1226


7alip (approved):
Thanks, very clear!

Issue #550, opened 2023-04-14
Manage roles and permissions through the dashboard.


MustafaTRKYLMZ:
- [ ] We can assign roles to users (update roles)
- [ ] create role
- [ ] update role
- [ ] delete role
- [ ] create user
- [ ] update user
- [ ] delete user

Örnek tasarım

![admin](https://user-images.githubusercontent.com/58182269/232084542-f9125c53-39b6-4af1-a7ca-b883e7065cbc.png)

PR #1238, opened 2024-04-24
Burak-role-viewer

myniqx:
In order to do what @MustafaTRKYLMZ  wanted, we first had to apply the detailed filtering that Strapi uses in its own API with the users-permissions plug-in to the frontend.

Therefore, in the first step, I added a list of all the APIs that the user can use along with his role to the profiles/me API. (you can check this permissions and play with them from this icon: 
 image

Then, I removed the fixed permission system and created the use of all these interactions, from the side menu items to be displayed in the system to the edit/delete buttons in the modals, dynamically, according to the APIs that come with the user's role.

so far it works.

I think I should make a presentation to you on Slack about these changes and then move on to the next steps.

https://github.com/freedomcombination/monorepo/issues/550


7alip:
@myniqx Can you update both branches with the latest dev?

PR #1239, opened 2024-04-24
Burak role viewer

myniqx:
In order to do what @MustafaTRKYLMZ  wanted, we first had to apply the detailed filtering that Strapi uses in its own API with the users-permissions plug-in to the frontend.

Therefore, in the first step, I added a list of all the APIs that the user can use along with his role to the profiles/me API. (you can check this permissions and play with them from this icon: 
 image

Then, I removed the fixed permission system and created the use of all these interactions, from the side menu items to be displayed in the system to the edit/delete buttons in the modals, dynamically, according to the APIs that come with the user's role.

so far it works.

I think I should make a presentation to you on Slack about these changes and then move on to the next steps.

https://github.com/freedomcombination/monorepo/issues/550


myniqx:
what im planing to do next;
1. step:
- first if you are admin, (your role's name admin) you ll have that icon (that image above) all the time.
- in the menu (when you click it) you can see all other roles, and fake switch between them to see what changes.
- while those changes if you are admin of course, navigation menu items ll not dissappear, they ll change color so we wont miss any.
2. step:
- save role from frontend
3. and so:
- design
- things about pages not directly releated to endpoints
- button component with endpoint and api props and that hide it self if there is no permission.. ( for edit / delete etc ) so we dont have to use permissionCheck functions all the time. if you didnt enter those props, lint ll alert.
- ...

packages/ui/tsconfig.json
```diff
@@ -1,5 +1,11 @@
 {
   "extends": "@fc/config/tsconfig.nextjs.json",
-  "include": ["next-env.d.ts", "@types/**/*.ts", "**/*.ts", "**/*.tsx"],
+  "include": [
+    "next-env.d.ts",
+    "@types/**/*.ts",
+    "**/*.ts",
+    "**/*.tsx",
+    "../types/src/permissions.ts"
```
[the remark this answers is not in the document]
7alip: Just revert thhis change :) VSCode sometimes adds it for some reason.

packages/context/src/auth/AuthProvider.tsx
```diff
@@ -36,6 +58,63 @@ export const AuthProvider: FC<AuthProviderProps> = ({
     }
   }, [initialState])
 
+  const permissionCheck = useMemo(() => {
+    const anyEndpoint = (endpoint: StrapiEndpoint): boolean => {
```
7alip, line 62:
```suggestion
    const hasPermission = (endpoint: StrapiEndpoint): boolean => {
```

packages/context/src/auth/AuthProvider.tsx
```diff
@@ -36,6 +58,63 @@ export const AuthProvider: FC<AuthProviderProps> = ({
...
+      return Object.values(permissions[makeSingular(endpoint)] || {}).some(
+        Boolean,
+      )
+    }
+
+    const fullEndpoint = (endpoint: StrapiEndpoint): boolean => {
```
7alip, line 68: Will we use it somewhere?

packages/context/src/auth/AuthProvider.tsx
```diff
@@ -36,6 +58,63 @@ export const AuthProvider: FC<AuthProviderProps> = ({
...
+      const name = makeSingular(endpoint)
+
+      return !permissions[name]
+        ? false
+        : Object.values(permissions[name] || {}).every(Boolean)
+    }
+
+    const apisEndpoint = (
+      endpoint: StrapiEndpoint,
+      ...apis: string[]
```
7alip, line 78: const hasActionPermission = (endpoint, ...actions)

packages/ui/src/admin/UserRoles/UserRoles.tsx
```diff
@@ -0,0 +1,138 @@
...
+  )
+
+  useEffect(() => {
+    const fetchRoles = async () => {
+      const list: SimpleRole[] = []
+      const { data: roleList } = await strapiRequest<Role>({
+        endpoint: 'users-permissions/roles',
+      })
+
+      for (const role of roleList) {
```
7alip, line 43: Can we create a new custom endpoint or is it possible to populate roles?

packages/ui/src/admin/UserRoles/UserRoles.tsx
```diff
@@ -0,0 +1,138 @@
...
+    'endpoint-filter',
+    [],
+  )
+
+  useEffect(() => {
+    const fetchRoles = async () => {
+      const list: SimpleRole[] = []
+      const { data: roleList } = await strapiRequest<Role>({
+        endpoint: 'users-permissions/roles',
+      })
```
7alip, line 41: token is missing

packages/ui/src/admin/UserRoles/UserRoles.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+      for (const role of roleList) {
+        const response = await strapiRequest<Role>({
+          endpoint: 'users-permissions/roles',
+          id: role.id,
+        })
```
7alip, line 47: token is missing

packages/context/src/auth/AuthProvider.tsx
```diff
@@ -36,6 +58,63 @@ export const AuthProvider: FC<AuthProviderProps> = ({
     }
   }, [initialState])
 
+  const permissionCheck = useMemo(() => {
```
7alip, line 61: Is there any performance issue when we don't group all these functions in `permissionCheck`? Couldn't we export all these functions individually?

7alip (commented):
I left some comments, thanks. I still have more reviewing to do I think.

packages/context/src/auth/AuthProvider.tsx
```diff
@@ -46,7 +125,16 @@ export const AuthProvider: FC<AuthProviderProps> = ({
         setUser(response.data?.user)
         setRoles(response.data?.user?.roles)
         setToken(response.data?.token)
-        setProfile(response.data?.profile)
+      }
+      const { permissions, ...profile } = response.data?.profile as Profile & {
+        permissions: StrapiPermission
```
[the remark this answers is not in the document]
myniqx: yes

packages/types/src/permissions.ts
```diff
@@ -0,0 +1,38 @@
+export type SimpleRole = {
```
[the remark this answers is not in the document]
myniqx: image
this Role's permissions is not a match.
strapi returns this structure for permission : 
image

packages/utils/src/permissions.ts
```diff
@@ -0,0 +1,75 @@
...
+  return Array.from(Object.keys(role.permissions))
+}
+
+/**
+ * Converts a Strapi permission object to a simplified endpoint object.
+ *
+ * @param {StrapiPermission} permission - The permission object to convert.
+ * @return {SimpleEndpoint} The converted simplified endpoint object.
+ */
+export const convertToSimple = (
```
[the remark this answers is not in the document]
myniqx: if ,in future, an undo (like map back) is not unnessary. i ll move it to backend.

packages/context/src/auth/AuthProvider.tsx
myniqx: i didnt think much naming, i accept :) but anyEndpoint looks if 'endpoint' has any api that allowed to user. anyEndpoint requires at least 1 api, fullEndpoint requires all.

packages/context/src/auth/AuthProvider.tsx
myniqx: i dont know yet :)

packages/types/src/permissions.ts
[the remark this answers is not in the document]
7alip: Yes I noticed it after the comment:

> Could we use this mapper in the backend and return the simplified permissions from api?

packages/context/src/auth/AuthProvider.tsx
7alip: I would remove it if it's not planned yet :)

packages/context/src/auth/AuthProvider.tsx
myniqx: i was tring to achive by that not recreate navItems in useAdminNav().. i only need it to recreate if permissions, language are changed. but when i added functions to dependency, it keeps rerender. i tried this way(useMemo) but somehow it didnt work to...

packages/context/src/auth/AuthProvider.tsx
7alip: `hasActionPermission` is like `boolean` variable name. Can be something like `checkActionPermission`

packages/context/src/auth/AuthProvider.tsx
7alip: Is it still a problem when you don't use useMemo and export the functions directly?

7alip: Well let's keep it like this. I'm just not sure if useMemo is necessary but maybe what you do is more performant.

myniqx: if i export functions and use them in dependency in useeffect/mome it does calculation in every single render. but since it still does, i can export them individually.

myniqx:
> I left some comments, thanks. I still have more reviewing to do I think.

after you finish reviewing can you let me know so i can do those changes and go next step.
and should i merge before i do that so you can track changes more easly?

packages/ui/src/admin/AdminNav/useAdminNav.tsx
```diff
@@ -27,208 +29,278 @@ import {
...
+    const setAllowedProps = (menuItem: AdminNavItemProps) => {
+      if (menuItem.allowed === true || menuItem.allowed === false) return
+
+      if (menuItem.submenu && menuItem.submenu.length > 0) {
+        menuItem.submenu.forEach(setAllowedProps)
+        menuItem.allowed = menuItem.submenu.some(submenu => submenu.allowed)
+      } else menuItem.allowed = allowByLink(menuItem.link)
+    }
+
+    menuItems.forEach(setAllowedProps)
```
7alip, line 290: Couldn't you use Array.map instead of modifying menuItems?

7alip:
```tsx
const mapAllowedMenu = (menuItem: AdminNavItemProps): AdminNavItemProps => {
  if (menuItem.allowed === true || menuItem.allowed === false) return menuItem;

  if (menuItem.submenu && menuItem.submenu.length > 0) {
    const submenu = menuItem.submenu.map(mapAllowedProps);
    const allowed = menuItem.submenu.some((submenu) => submenu.allowed);

    return { ...menuItem, submenu, allowed };
  } else {
    const allowed = allowByLink(menuItem.link);

    return { ...menuItem, allowed };
  }
};

return menuItems.map(mapAllowedMenu).filter(filterMenu)
```

apps/api/src/api/profile/controllers/custom.ts
```diff
@@ -24,13 +26,30 @@ module.exports = {
...
 
+      const roleId = ctx.state.user.role.id
+      const roleResponse =
+        await strapi.plugins['users-permissions'].services.role.findOne(roleId)
+
       const newProfile = { ...profile }
       delete newProfile.user
 
-      return { data: profile }
+      return { data: { ...profile, permissions: roleResponse.permissions } }
```
7alip, line 36: @myniqx Why don't we mapPermissions here but in `getRoles`? This way, we use `mapPermissions` both in frontend and backend.

myniqx: This is not final work. I mentioned that i havent checked what kind an object i need to update a role. In order me be able to pass next step i made those suggested changes and waiting. I may delete mapPermissions if it become redundant.

7alip: It's not a big deal, but it would be nice to have the mapper only in the frontend or in the backend.

Perhaps it is time to investigate how you will update the permissions from the client. Is there an existing endpoint for it, or do you need to create custom controllers? Which parameters will you need, etc.

7alip: You can merge this into your actual branch if you want. I don't have any idea yet about how to handle custom dashboard pages. We can discuss it later.

PR #1251, opened 2024-05-02
Burak role viewer part - 2

myniqx:
i think i come somewhere.
all functions are working so far.


MustafaTRKYLMZ:
image
My left side bar is like this. Is it normal? @myniqx

MustafaTRKYLMZ:
Some permissions must open to public role in Dashboard page
1-News

- [ ] News (read and bookmark only)
- [ ] Bookmarked news (CRUD)

2- Timeline 

- [ ] Tweets (read and bookmark only)
- [ ] Bookmarked tweets (CRUD)

3- Donation (users can donate)(It is not related with users who donated)(It is like public role)

4-Feeback

- [ ] Create feedback

Issue #780, opened 2023-08-08
Add activities and blogs to recommended news page


MustafaTRKYLMZ:
### Tasks

- [ ] add activities and blogs to recommended news page
- [ ] Show last month data (hide delete button)

PR #1252, opened 2024-05-03
Burak-news

myniqx:
https://github.com/freedomcombination/monorepo/issues/780


apps/trend-rights/pages/api/images.ts
```diff
@@ -0,0 +1,7 @@
+import { imagesRouter } from '@fc/services/src/api/images'
```
7alip, line 1: Is this API route being used?

7alip (commented):
image

I discussed with @MustafaTRKYLMZ and it's probably cleaner not to combine topics in one page but show blogs and activities in different tabs. They will still be displayed in Topic card the same way you implemented.

You can also use only client side fetch and add skeletons while fetching. I would create custom hooks for it.

An example approach

```tsx
const useMappedBlogs = () => {
  const { data, ...blogsQuery } = useStrapiRequest<Blog>()
  
  const mappedBlogTopics = data?.data?.map()
  
  return { data: mappedBlogTopics, ...blogsQuery }
}
```

Each TabContent can its own component. Since you use localStorage in Activity and Blog components, they can be imported dynamically in the client.

Chakra tab contents are loaded dynamically. If you fetch the content inside the tab content components, the fetch will be trigger only the tab is active.

apps/trend-rights/pages/news.tsx
```diff
@@ -152,24 +177,142 @@ export const getServerSideProps = async (
...
-  }))
+
+  const topics = queryClient.getQueryData<TopicBase[]>(['topics']) || []
 
   return {
     props: {
       seo,
       topic: recommendedTopic as RecommendedTopic | null,
-      topics: recommendedTopics,
+      topics,
```
7alip, line 191: I noticed that we don't use this page for SEO purposes. When we share a news card, we point to its original source.

For blogs and activities, we can point to their original pages in foundation website. In this case we don't have to do server side fetch and `topic` is redundant here.

myniqx:
this is final state:
image

apps/trend-rights/pages/api/images.ts
7alip: @myniqx I'm wondering where this API and world-map.svg are used?

myniqx: TopicCard -> backgroundImage
image

myniqx: and API, when user clicks to create post ... it leads to crush. that is when i realize this.
CreateModelForm -> 
image
and useFileFormUrl ->
image

7alip: Thanks, I just forgot what I did in the past :)

packages/ui/src/admin/TopicCard/TopicCard.tsx
```diff
@@ -8,6 +8,7 @@ import {
   Popover,
   PopoverArrow,
   PopoverBody,
+  Highlight,
```
7alip, line 11: Great! I didn't know that :)

7alip (approved):
Great job! Thanks

PR #1255, opened 2024-05-05
Burak role viewer

myniqx:
since lastnight;
added Action<Component>s like ActionButton etc.
fixed navbars items
   - all items are visible
   - no flickering while authCheck
change all conditional rendering to ActionComponents..

could you both check, please, if there is a bug left, my brain is not braining right now.

ps: after merge, roles has to be set on production database and `yarn db` is a good start for all other developers


7alip:
@myniqx Build error message: "The lockfile would have been modified by this install, which is explicitly forbidden.". You should run `yarn install` and push it after committing the updated `yarn.lock`

7alip (approved):
@myniqx Let's merge this into dev. I'll prepare the staging API then we can test it in staging before merging into production.

We can create a followup task if needed

PR #1266, opened 2024-05-08
status field added to strapi

MustafaTRKYLMZ:
Volunteer status workflow

**=> Pending:** This status indicates that an applicant has applied to become a volunteer, but no decision has been made yet.

**=> Approved:** After the first meeting, if the applicant is deemed suitable for further consideration, they are moved to the approved status. This means they have passed the initial evaluation.

**=> Accepted:** Following a second meeting where everything is satisfactory, the applicant becomes accepted, meaning they are officially welcomed as a volunteer.

**=> Rejected:** If for any reason the organization decides not to proceed with a volunteer, or if the volunteer chooses not to continue, they are marked as rejected.

**=> In Progress:** This status is assigned when a meeting is planned, but it hasn't taken place yet. It signifies that the application process is actively moving forward.

**=> Awaiting:** After a meeting has occurred but a decision hasn't been reached yet, the status is changed to awaiting. This could be due to needing further deliberation or additional information.

**=> Left:** If a volunteer decides to withdraw from the program for any reason, they are marked as left. This helps in tracking volunteer turnover and understanding why volunteers choose to leave.


packages/types/src/profile.ts
```diff
@@ -22,6 +22,15 @@ import { User } from './user'
...
 import { Vote } from './vote'
 
+export type ProfileStatus =
+  | 'left'
+  | 'pending'
+  | 'rejected'
+  | 'in-progress'
+  | 'accepted'
+  | 'awaiting'
+  | 'approved'
```
myniqx, line 32: > Approved: After the first meeting, if the applicant is deemed suitable for further consideration, they are moved to the approved status. This means they have passed the initial evaluation.

Maybe those informations can be added as comment.

myniqx (approved):

7alip (approved):

PR #1274, opened 2024-05-11
platform and profilestatus added

MustafaTRKYLMZ:
For profiles page
Platform and profileStatus filter added


7alip (approved):
I create a new PR with some design improvements. You can merge both if there is no issue

PR #1284, opened 2024-05-12
Burak-uploader

myniqx:
this is how its look like:
image

if you havent upload any image in this session, you can always get images from server ( image ids are saved in local storage and show them when you click this option)
image

in list, you can see a preview and file formats. you can delete image or select a file format to paste code into markdown editor.
image

after you select a file format, it copies image code (image) to clipboard. when set a cursor to markdown editor and paste, it looks like this :
image


MustafaTRKYLMZ:
It is looking so perfect. But I am wondering that it is possible to upload other format at the same time such as pdf, exel and doc.

7alip (approved):
Thanks

packages/ui/src/components/FormUploader/MenuImageItem.tsx
```diff
@@ -0,0 +1,121 @@
...
+  children: ReactNode
+}
+
+const MenuImageButton: FC<MenuImageButtonProps> = ({ url, children }) => {
+  const { onCopy, setValue, hasCopied } = useClipboard('')
+
+  useEffect(() => {
+    // TODO: Change it to link if it's not an image
+    setValue(`![image](${API_URL}${url})`)
+  }, [url])
```
7alip, line 37: This comment can be removed after fixing

packages/ui/src/components/FormUploader/MenuImageItem.tsx
```diff
@@ -0,0 +1,121 @@
...
+    </Button>
+  )
+}
+
+export const MenuImageItem: FC<MenuImageItemProps> = ({ image, onDelete }) => {
+  const entries = Object.entries(image.formats ?? {})
+
+  return (
+    <Box pos={'relative'}>
+      {/* TODO: Display pdf icon for pdfs, File icon for other file types */}
```
7alip, line 70: This comment can be removed after fixing.

Icon wrapper box can have the same height as image.

7alip (changes requested):
Thanks

myniqx:
file icons:
image

it pastes as link when its not an image:
image

7alip:
@myniqx The only thing left is to remove dimensions part if with or height is not available. It shows only `x` now

myniqx:
> @myniqx The only thing left is to remove dimensions part if with or height is not available. It shows only `x` now

```
           <VStack fontSize={'xs'} spacing={0}>
              <Box>original</Box>
              {isImage ? (
                <Box>
                  {file.width} x {file.height}
                </Box>
              ) : (
                <Box>{(file.size / 1024).toFixed(2)} kb</Box>
              )}
            </VStack>
```
its already removed. if file is image it ll show dimensions otherwise it ll show file size.
and beside image file they dont have fileformats so there ll be only this 'original' box with filesize, u can see in this image : 
image

7alip (approved):
Thanks

PR #1285, opened 2024-05-12
fetched foundationJobs in join page

ertugrul-aydin:


ertugrul-aydin:
How can we fetch jobs related to the necessary filtering feature when pulling jobs from Strapi?

packages/ui/src/templates/JoinTemplate/JoinTemplate.tsx
```diff
@@ -44,6 +44,14 @@ export const JoinTemplate: FC<JoinTemplateProps> = ({ title }) => {
 
   const platforms = platformsResult.data?.data || []
 
+  // Fetch foundation jobs
+  const foundationJobsResult = useStrapiRequest<Job>({
+    endpoint: 'jobs',
+    locale,
```
7alip, line 50: @ertugrul-aydin Did you try filtering jobs by platform here?

- `filters: { platform: { $null: true } }` or
- `filters: { platform: { $eq: null } }`

ertugrul-aydin: I noticed there were some jobs in Strapi without any platform linked to them, so the filter wasn't working. Now, after adding those jobs without any platform linked, they're showing up under Foundation.

packages/ui/src/components/JoinForm/JoinForm.tsx
```diff
@@ -198,6 +199,23 @@ export const JoinForm: FC<JoinFormProps> = ({
...
+                key={job.id}
+                id={job.id.toString()}
+                {...register(`jobs`)}
+                value={job.id}
+                textTransform={'capitalize'}
+              >
+                {job[`name_${locale}`]}
+              </Checkbox>
+            </Stack>
+          ))}
```
7alip, line 217: Wrap this section with `Stack` like other platform jobs section and show only when there are foundationJobs:

```tsx
{foundationJobs.length > 0 && <Stack ...}
```

MustafaTRKYLMZ: We don't have any relation between jobs and foundation. Is it fixed here or not? one to many -maybe

7alip: @MustafaTRKYLMZ If a job has no platform relation, we assume that it is related to the foundation.

MustafaTRKYLMZ: Definitly, For example network coordinator (state members) don't have any platform.

MustafaTRKYLMZ: > Definitly, For example network coordinator (state members) don't have any platforms.

It means we have to add foundation (it must be optional) relations to the profiles.

7alip (approved):
Thanks

7alip:
@ertugrul-aydin This can be merged

PR #1297, opened 2024-05-15
Edit observation

MustafaTRKYLMZ:
Observation edit and delete added


packages/ui/src/admin/ObservationEditForm/ObservationEditForm.tsx
```diff
@@ -0,0 +1,152 @@
...
+        deleteMutation.mutate({ id }, { onSuccess: handleSuccess })
+        setConfirmState(undefined)
+      },
+    })
+  }
+
+  const onSave = async (data: ObservationUpdateInput) => {
+    const body = {
+      content: data.content,
+    } as ObservationUpdateInput
```
7alip, line 86: In typescript `as` can force type and normally it's not always recommended. You can either use the followings:

```ts
const body: ObservationUpdateInput = {
  content: data.content,
}
```

or, `satisfies` instead of `as`

```ts
const body = {
  content: data.content,
} satisfies ObservationUpdateInput
```

If you do that, you will notice that only having `content` doesn't satisfy the type. Because `ObservationUpdateInput` has `profile` and `creator`. We set them in the backend and they can be removed from `ObservationUpdateInput`.

packages/types/src/observation.ts
```diff
@@ -19,5 +19,8 @@ type ObservationRelationInput = {
 export type ObservationCreateInput = Expand<
   Pick<ObservationBase, 'content'> & ObservationRelationInput
 >
+export type ObservationUpdateInput = Expand<
+  Partial<ObservationBase> & ObservationRelationInput
+>
```
7alip, line 24:
```tsx
export type ObservationCreateInput = Pick<ObservationBase, 'content'>

export type ObservationUpdateInput = ObservationCreateInput
```

7alip (changes requested):
Thanks

7alip (approved):
Thanks

Issue #1279, opened 2024-05-11
Replace `Generate Posts` with ArchiveContents edit form


7alip:
### Is your feature request related to a problem? Please describe.

In dashboard, we use `/archive-contents` to display the table of ArchiveContents. But in the detail page (`/archive-content/:id`), we display **Post Generator** instead of model edit modal.

### Describe the solution you'd like

- [ ] Display ArchiveContent edit form on the `/archive-content/:id` page.
- [ ] Add `Generate Posts` button if the endpoint is `archive-content` in edit form.
- [ ] `Generate Posts` button should redirect users to `/generate-post/:archive-content-id` page.

### Describe alternatives you've considered

No considered alternatives

### Additional context

No additional context

PR #1304, opened 2024-05-16
Burak-archive-content

myniqx:
https://github.com/freedomcombination/monorepo/issues/1279

-  Display ArchiveContent edit form on the `/archive-content/:id` page.
this part solved by using [endpoint].tsx file.

-  Add `Generate Posts` button if the endpoint is `archive-content` in edit form.
this part solved bu adding extra to ActionButton
```
            <ActionButton
              isVisible={
                endpoint === 'hashtags' || endpoint === 'archive-contents'
              }
              checkActions={{ endpoint: 'posts', actions: ['createPosts'] }}
              onClick={() => {
                if (endpoint === 'hashtags') router.push(`/hashtags/${id}`)
                else router.push(`/archive-contents/${id}`)
              }}
              leftIcon={<FaXTwitter />}
              fontSize="sm"
              colorScheme={'purple'}
              isLoading={approveModelMutation.isPending}
            >
              {t('posts')}
            </ActionButton>
```
now, same button when editing hashtag that appears will appear while editing archive-contents as well.

- `Generate Posts` button should redirect users to `/generate-post/:archive-content-id` page.
and as you can see, i avoid to create new route 'generate-post' because it ll not be good with permission system, its a new route and there is no admin navigator link starts with generate-post so it ll be rejected all the time.
this is how it works now:
`/archive-contents?id=4`  that means [endpoint].tsx ll work with id 4 which is editing page.
`/archive-contents/4` means arhive-contents/:id ll work which is old generate post page.


packages/ui/src/admin/ModelForm/ModelEditForm.tsx
```diff
@@ -492,9 +492,14 @@ export const ModelEditForm = <T extends StrapiModel>({
...
-              isVisible={endpoint === 'hashtags'}
+              isVisible={
+                endpoint === 'hashtags' || endpoint === 'archive-contents'
+              }
               checkActions={{ endpoint: 'posts', actions: ['createPosts'] }}
-              onClick={() => router.push(`/hashtags/${id}`)}
+              onClick={() => {
+                if (endpoint === 'hashtags') router.push(`/hashtags/${id}`)
+                else router.push(`/archive-contents/${id}`)
+              }}
```
7alip, line 502: I would move it to a function

```suggestion
              onClick={onPostsClick}
```

7alip (approved):

Issue #1166, opened 2024-04-08
Add recaptcha to the feedback form


7alip:
Recaptcha protection has been implemented for art and blog-like actions, as well as the post maker share button. Use the same method for the feedback form.

See how `ReCaptchaProvider`, `useRecaptchaToken` are used in the client side and `checkRecaptcha` in the server.

PR #1305, opened 2024-05-16
Burak-feedback

myniqx:
https://github.com/freedomcombination/monorepo/issues/1166

can you check in your system please, because recaptcha allways (in everywhere) returns my request.

### DIFF
--- a/packages/services/src/userFeedback/userFeedback.ts
+++ b/packages/services/src/userFeedback/userFeedback.ts
@@ -8,6 +8,7 @@ import { UserFeedback, UserFeedbackCreateInput } from '@fc/types'
 export const createUserFeedback = async (
   userFeedback: UserFeedbackCreateInput,
   token: string,
+  recaptchaToken?: string,
 ) => {
   if (!userFeedback) {
     throw new Error('feedback field is required')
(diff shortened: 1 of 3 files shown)


packages/services/src/userFeedback/userFeedback.ts
```diff
@@ -8,26 +8,30 @@ import { UserFeedback, UserFeedbackCreateInput } from '@fc/types'
...
     throw new Error('feedback field is required')
   }
 
   await Mutation.post<UserFeedback, UserFeedbackCreateInput>(
     'user-feedbacks',
-    userFeedback,
+    {
+      ...userFeedback,
+      recaptchaToken,
+    } as UserFeedbackCreateInput,
```
7alip, line 22: _You don't have to change anything, it's just a note._

New TS operator `satisfies` is more strict and recommended.

```suggestion
    } satisfies UserFeedbackCreateInput,
```

7alip (approved):
Thanks

packages/services/src/userFeedback/userFeedback.ts
myniqx: i remembered. but I specifically used 'as'. because when recaptchaToken was included, it is no longer satisfied the UserFeedbackCreateInput object. I made it so ts would ignore it.

PR #1322, opened 2024-05-19
Refactor push notifications

7alip:


packages/ui/src/components/SubToNotification/SubToNotification.tsx
```diff
@@ -13,43 +13,54 @@ export const SubToNotification = () => {
     useState<ServiceWorkerRegistration | null>(null)
 
   useEffect(() => {
-    if (
-      typeof window !== 'undefined' &&
-      'serviceWorker' in navigator &&
-      (window as any).workbox !== undefined
-    ) {
+    const registerServiceWorker = async () => {
+      if (!('serviceWorker' in navigator) || !('workbox' in window)) {
```
7alip, line 17: Double check if it could be `PushManager` instead of `workbox`

```suggestion
      if (!('serviceWorker' in navigator) || !('PushManager' in window)) {
```

PR #1331, opened 2024-05-21
Feat/logs

myniqx:
image


apps/api/src/libs/subscribeDb.ts
```diff
@@ -0,0 +1,108 @@
...
+    if (
+      event.action !== 'afterDelete' &&
+      /*
+         event.action !== 'afterUpdate' &&
+
+         - beforeUpdate has params.data which includes the keys that ll be updated
+          so we can understand if it's for approval or publishing
+      */
+      event.action !== 'afterCreate' &&
+      event.action !== 'beforeUpdate'
```
7alip, line 55: Maybe we can do the same way as allowedApis.

`const allowedActions = []`

packages/ui/src/admin/AuditLogTable/AuditLogTable.tsx
```diff
@@ -0,0 +1,101 @@
+import React, { FC, useState } from 'react'
```
7alip, line 1: We can consider a different design for logs but for now lets display in the table view

7alip (approved):
Thanks

PR #1332, opened 2024-05-21
Clean user stats

7alip:


myniqx (approved):
After the log issue, the event listening function with 'subscribe' may be useful here as well.

- As you did in that issue, you can make a uid list and fill in the approver or creator sections after afterCreate.
- If there is an event.model object in afterCreate event like beforeUpdate event, that object indicates the presence or absence of the approver or creator columns of the relevant model. a single function can be used to support all models to be added in the future, without even needing to list uids.

In this way, we can follow a path that is easy to modify when necessary, from a single place.

PR #1334, opened 2024-05-21
Update audit logs design

7alip:


apps/api/src/libs/subscribeDb.ts
```diff
@@ -83,13 +89,7 @@ export const subscribeDb = async () => {
           result?.name_en ||
           result?.name_tr ||
           result?.name_nl ||
-          result?.username ||
-          data?.approvalStatus ||
```
myniqx, line 87: ps: removing data ll result as blank text. because result is not exist beforeUpdate.

packages/ui/src/admin/AuditLogList/AuditLogList.tsx
```diff
@@ -0,0 +1,164 @@
...
+import { addDays, format } from 'date-fns'
+
+import { useAuthContext } from '@fc/context'
+import { strapiRequest } from '@fc/lib'
+import { AuditLog } from '@fc/types'
+
+import { AuditLogItem } from './AuditLogItem'
+import { SearchForm } from '../../components'
+
+export const AuditLogList: FC = () => {
```
myniqx, line 14: just you to see how it looks like when logs appear:
image

myniqx (approved):

apps/api/src/libs/subscribeDb.ts
7alip: Oh, I completely forgot it

packages/ui/src/admin/AuditLogList/AuditLogList.tsx
7alip: It's probably because of your existing AuditLogs data. I updated actions.

PR #1348, opened 2024-05-26
Update subscribeDb

7alip:


apps/api/src/libs/subscribeDb.ts
```diff
@@ -58,44 +83,101 @@ export const subscribeDb = async () => {
...
+      data?.name_nl ||
+      data?.username ||
+      data?.description ||
+      data?.description_en ||
+      data?.description_tr ||
+      data?.description_nl ||
+      data?.content ||
+      data?.content_en ||
+      data?.content_tr ||
+      data?.content_nl ||
```
myniqx, line 152: the content of many models are richtext, and audit's text is only text which has limit of 255 character.
it happend to me, it ll cause crash.

apps/api/src/libs/subscribeDb.ts
```diff
@@ -58,44 +83,101 @@ export const subscribeDb = async () => {
...
-          ? 'approved'
-          : data.approvalStatus === 'rejected'
-            ? 'rejected'
-            : data.publishedAt
-              ? 'published'
-              : 'updated'
-    } else if (event.action === 'afterDelete') {
-      action = 'deleted'
+    if (locale) {
+      text = `${text} (${locale})`
```
myniqx, line 174: we can make sure here that text has 255 char max.

myniqx (approved):

apps/api/src/libs/subscribeDb.ts
7alip: Good catch! I used to add `.slice()` but after some refactoring I forgot to add it back. I'm not sure if we use markdown editor or richtext but it's probably best to include only if the content is string type

myniqx: additionally: 
``` 
      data?.title ||
      data?.title_en ||
      data?.title_tr ||
      data?.title_nl ||
      data?.name ||
      data?.name_en ||
      data?.name_tr ||
      data?.name_nl ||
      data?.username ||
      data?.description ||
      data?.description_en ||
      data?.description_tr ||
      data?.description_nl ||
      data?.content ||
      data?.content_en ||
      data?.content_tr ||
      data?.content_nl || 
``` 

it may never happen too. because we never update solo `content`. those `or`s ll return `title` most of the time.

7alip: I noticed that we nothing was displayed after only the content field had been updated. Perhaps it was from the Strapi dashboard. That is why I included them just in case.

PR #1357, opened 2024-05-27
update profile if already exist

myniqx:
im not sure if we check name in query
` $and: [{ email: { $eq: email } }, { **name: { $eq: name }** }],`

### DIFF
--- a/apps/api/src/api/profile/controllers/profile.ts
+++ b/apps/api/src/api/profile/controllers/profile.ts
@@ -37,14 +37,14 @@ export default factories.createCoreController('api::profile.profile', () => {
     async create(ctx) {
       if (ctx.request.body?.data?.recaptchaToken) await checkRecaptcha(ctx)
 
-      const { user: userId, name, email } = ctx.request.body.data
+      const { email } = ctx.request.body.data
 
       // we dont want to use find here because it has modified...
       const profileResults = await strapi.db
         .query('api::hashtag.hashtag')
         .findMany({
           where: {
-            $and: [{ email: { $eq: email } }, { name: { $eq: name } }],
+            $and: [{ email: { $eq: email } }],
           },
         })
 
@@ -57,7 +57,7 @@ export default factories.createCoreController('api::profile.profile', () => {
 
       await strapi.entityService.update('api::profile.profile', profile.id, {
         data: {
-          user: userId,
+          user: ctx.state?.user?.id,
         },
       })
     },
--- a/apps/api/src/api/profile/controllers/profile.ts
+++ b/apps/api/src/api/profile/controllers/profile.ts
@@ -37,7 +37,29 @@ export default factories.createCoreController('api::profile.profile', () => {
     async create(ctx) {
       if (ctx.request.body?.data?.recaptchaToken) await checkRecaptcha(ctx)
 
-      return super.create(ctx)
+      const { user: userId, name, email } = ctx.request.body.data
+
+      // we dont want to use find here because it has modified...
+      const profileResults = await strapi.db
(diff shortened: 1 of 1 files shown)


apps/api/src/api/profile/controllers/profile.ts
```diff
@@ -37,7 +37,29 @@ export default factories.createCoreController('api::profile.profile', () => {
     async create(ctx) {
       if (ctx.request.body?.data?.recaptchaToken) await checkRecaptcha(ctx)
 
-      return super.create(ctx)
+      const { user: userId, name, email } = ctx.request.body.data
```
7alip, line 40: I think we can avoid passing `user` in the body because we can read it from `ctx.state.user`.

In the frontend,

```ts
const body: ProfileCreateInput = {
  user: userId, // Remove it
  name: trimmedName,
  email: trimmedEmail,
};

await Mutation.post("profiles", body, response.data?.jwt);
```

and get user info from the state.

```ts
const { name, email } = ctx.request.body.data
const userId = ctx.state?.user?.id
```

apps/api/src/api/profile/controllers/profile.ts
```diff
@@ -37,7 +37,29 @@ export default factories.createCoreController('api::profile.profile', () => {
...
+
+      // we dont want to use find here because it has modified...
+      const profileResults = await strapi.db
+        .query('api::hashtag.hashtag')
+        .findMany({
+          where: {
+            $and: [{ email: { $eq: email } }, { name: { $eq: name } }],
+          },
+        })
```
7alip, line 49: 1. Do you mean `strapi.entityService.find` here?. I think you trigger the controller from another code only when you call `strapi.service(uid)` not entityService.

2. The email field should have a unique constraint, and I don't think we should include `name` here. Because they can modify their name in the future from the profile page.

7alip: You can also try this if it works

```ts
strapi.db.query("api::profile.profile").findOne({
  where: {
    email: { $eq: email },
  },
});
```

7alip (commented):
Thanks

apps/api/src/api/profile/controllers/profile.ts
myniqx: `user: userId, // Remove it` 
are you sure? it is needed to create profile from super.create ? we shouldnt remove it from frontend to use super.create.

apps/api/src/api/profile/controllers/profile.ts
myniqx: said gemini;

>In short, functions like find that you override in your Strapi model will always call that Strapi function. That is, it does not query the database directly via db.query.

even db.queries which are low level query functions than entityService use that overrided functions.

apps/api/src/api/profile/controllers/profile.ts
7alip: Aha, it's needed indeed! Thanks.

apps/api/src/api/profile/controllers/profile.ts
```diff
@@ -37,7 +37,29 @@ export default factories.createCoreController('api::profile.profile', () => {
...
 
-      return super.create(ctx)
+      const { email } = ctx.request.body.data
+
+      // we dont want to use find here because it has modified...
+      const profileResults = await strapi.db
+        .query('api::hashtag.hashtag')
+        .findMany({
+          where: {
+            $and: [{ email: { $eq: email } }],
```
7alip, line 47: `$and` is not necessary anymore

7alip (approved):
Thanks

PR #1371, opened 2024-06-02
Refactor blog findOne controller

7alip:
@ertugrul-aydin Can you do some tests if this works


apps/api/src/api/blog/controllers/blog.ts
```diff
@@ -4,36 +4,30 @@ import { factories } from '@strapi/strapi'
...
-
       const isLiked =
         user &&
         (await strapi.db.query('api::blog.blog').count({
           where: {
-            slug: searchValue,
-            likers: {
-              email: user.email,
-            },
+            slug,
```
7alip, line 19: this should also be `id ? { id } : { slug }`

PR #1387, opened 2024-06-04
Send email from profile

MustafaTRKYLMZ:


7alip (changes requested):
https://github.com/freedomcombination/monorepo/pull/1387/files

I don't see any changes in this PR. And it seems the source branch is created based on main not dev

MustafaTRKYLMZ:
> https://github.com/freedomcombination/monorepo/pull/1387/files
> 
> I don't see any changes in this PR. And it seems the source branch is created based on main not dev

It is possible to see changes in last commit. Main was behind of the my branch. Reason is that.

PR #1391, opened 2024-06-05
observation create added after email sended

MustafaTRKYLMZ:


packages/ui/src/admin/ProfileEmailForm/ProfilEmailForm.tsx
```diff
@@ -46,20 +57,57 @@ export const ProfileMailForm: FC<ProfileMailFormProps> = ({ email }) => {
...
-        subject: data.subject,
-        text: data.content,
+        subject,
+        text: content,
       }
 
       await sendEmail(emailData)
       reset()
+
+      isSuccess && createObservation({ subject, content })
```
7alip, line 73: The function `createObservation` is being used before it's defined. While modern compilers might handle this, it's best practice to define functions before they're used to avoid potential issues.

There are two ways to define a function: **function declaration** and **function expression**.


```ts
// Function Declaration
function createObservation() {}
```

With function declarations, the function can be used before it's defined due to **hoisting**.

```ts
// Function Expression:
const createObservation = () => {}
```

However, with function expressions, the function must be defined before it's used. This is because the variable createObservation is hoisted but its assignment as a function happens at runtime.

So, if you're using a function expression, ensure createObservation is defined before it's used.

---

**Problem due to hoisting**
```ts
myFunction() // Don't use before it's defined

const myFunction = () => {}
```

**Not problem**
```ts
myFunction() // Can be used if it's defined as function expression

function myFunction() {}
```

packages/ui/src/admin/ProfileEmailForm/ProfilEmailForm.tsx
```diff
@@ -46,20 +57,57 @@ export const ProfileMailForm: FC<ProfileMailFormProps> = ({ email }) => {
...
+  type createOnservationProps = {
+    subject: string
+    content: string
+  }
+
+  const createObservation = async ({
+    subject,
+    content,
+  }: createOnservationProps) => {
+    const observationContent = `email sended to: ${email} \n with subject: ${subject} \n and content: ${content}`
```
7alip, line 93:
```suggestion
    const observationContent = `Email sent to: ${email}\n\n with subject: ${subject}\n\n and content: ${content}`
```

packages/ui/src/admin/ProfileEmailForm/ProfilEmailForm.tsx
```diff
@@ -46,20 +57,57 @@ export const ProfileMailForm: FC<ProfileMailFormProps> = ({ email }) => {
...
+
+    try {
+      const body = {
+        content: observationContent,
+        profile: profileId,
+      } as ObservationCreateInput
+
+      mutate(body, { onSuccess: () => onSuccess?.() })
```
7alip, line 101: Can also be

```suggestion
      mutate(body, { onSuccess })
```

packages/ui/src/admin/ProfileEmailForm/ProfilEmailForm.tsx
```diff
@@ -46,20 +57,57 @@ export const ProfileMailForm: FC<ProfileMailFormProps> = ({ email }) => {
...
     try {
       const emailData: EmailCreateInput = {
         to: email,
-        subject: data.subject,
-        text: data.content,
+        subject,
+        text: content,
       }
 
       await sendEmail(emailData)
```
7alip, line 70: Using onSuccess callback is probably better

```ts
      await sendEmail(emailData, {
        onSuccess: () => createObservation({ subject, content }),
      })
      reset()
```

7alip (commented):
Thanks

7alip (approved):
Thanks

PR #1392, opened 2024-06-05
Burak profile page

myniqx:
profile page added as modal to dashboard:

Ekran görüntüsü 2024-06-05 125517

it can be customized according to app:
Ekran görüntüsü 2024-06-05 125732

arts tab is not available to all apps:
Ekran görüntüsü 2024-06-05 130010

those screenshots are not final, i mean when you run 'yarn dev' you can see diffrent komponents
Ekran görüntüsü 2024-06-05 130520

i added an bash script to add entry to i18 new word:
`yarn add-word key value` this command adds new values to all tr,nl,en/common.json (only in packages/ui/public)
`yarn copy-locales` copies all files in package/ui/public/locale to other applications...
there were too many entries. for convenience in the future, I also created a component to edit these entries. it works only development mode. you can see that in dashboard > translate.

there are 2 actions. /profile/user (this is new and used to change username and email) and changepassword... all roles has to have those 2 action's permission.


apps/api/src/api/profile/controllers/custom.ts
```diff
@@ -47,4 +47,43 @@ module.exports = {
...
+        userId,
+        {
+          data: {
+            email,
+          },
+        },
+      )
+    }
+
+    throw new Error('No data provided')
```
7alip, line 87: In case of updating the username and email at the same time, wouldn't it be nicer to run the update function just once?

```ts
const username = data.username
const email = data.email

if (!data || (!username && !email)) {
  // You can return strapi validation errors from @strapi/utils
  throw new Error('No data provided')
}

return await strapi.entityService.update(
  'plugin::users-permissions.user',
  userId,
  {
    data: {
      ...(username && { username }),
      ...(email && { email }),
    },
  },
)
```

apps/dashboard/next.config.mjs
```diff
@@ -2,8 +2,12 @@ import withPWAInit from '@ducanh2912/next-pwa'
 
 import nextConfig from '@fc/config/next.config.mjs'
 
-const withPWA = withPWAInit({
-  dest: 'public',
-})
+const withPWA = () => {
```
7alip, line 5: Do we still need this?

7alip (commented):
Excellent job! It would be nice to review each feature in a separate PR but that's ok :)

I'll adjust stylings later.

apps/api/src/api/profile/controllers/custom.ts
myniqx: in frontend, there is no way update them once.

PR #1395, opened 2024-06-06
refresh observation after the sent email: fixed

MustafaTRKYLMZ:
There was an error when creating observation after sent email... I think It was about refreshing observations. 

First of all, creating observation is successful, but the warning say: "Something went wrong"


apps/dashboard/pages/[endpoint].tsx
```diff
@@ -150,6 +151,7 @@ const ModelPage: FC<ModelPageProps> = ({ endpoint }) => {
       enabled: !!endpoint && !!token,
     },
   })
+  // feth observation
```
Comment (7alip) [inline apps/dashboard/pages/[endpoint].tsx:154]: Comment can be removed

7alip (approved):
Thanks

PR #1397, opened 2024-06-06
Refactor profile

MustafaTRKYLMZ:


packages/ui/src/admin/ProfileEmailForm/ProfilEmailForm.tsx
```diff
@@ -68,7 +68,7 @@ export const ProfileMailForm: FC<ProfileMailFormProps> = ({
     subject,
     content,
   }: createOnservationProps) => {
-    const observationContent = `Email sent to: ${email}\n\n with subject: ${subject}\n\n and content: ${content}`
+    const observationContent = `Email sent to: ${email}<br><br> with subject: ${subject}<br><br> and content: ${content}`
```
7alip, line 71: It's not necessary to add `<br />` to observation content as it's plain long text field.

packages/ui/src/admin/ProfileEmailForm/ProfilEmailForm.tsx
```diff
@@ -90,7 +90,7 @@ export const ProfileMailForm: FC<ProfileMailFormProps> = ({
   const { error, isPending, isSuccess, mutateAsync: sendEmail } = useSendEmail()
 
   const onSubmit = async (data: EmailFormValues) => {
-    const content = data.content
+    const content = data.content.replace(/\n/g, '<br>')
```
7alip, line 93: This is a temporary workaround. We will use richtext editor or templates in the future.

7alip (changes requested):

packages/ui/src/admin/ProfileEmailForm/ProfilEmailForm.tsx
MustafaTRKYLMZ: > It's not necessary to add `<br />` to observation content as it's plain long text field.

I thought it would be more readable if we split the sentence into two or three lines.

Issue #1363, opened 2024-05-28
Enhance push notifications (dynamic content)


hakanayata:
This is a follow-up to issue #618 

- [x] Make push notification content dynamic

PR #1411, opened 2024-06-09
Dynamic notification content

hakanayata:
This PR addresses issue #1363


packages/ui/src/admin/ModelForm/ModelEditForm.tsx
```diff
@@ -302,6 +303,25 @@ export const ModelEditForm = <T extends StrapiModel>({
...
+  > = async event => {
+    event.preventDefault()
+    if (!defaultValues) return
+
+    try {
+      await fetch('/api/notification', {
+        method: 'POST',
+        body: JSON.stringify({
+          title: defaultValues.title,
+          message: defaultValues.body,
```
7alip, line 317: Are we able to change the title and body fields from the form? Shouldn't they be read from `watch`?

```ts
const [title, body] = watch(['title', 'body']
```

packages/ui/src/admin/ModelForm/ModelEditForm.tsx
```diff
@@ -511,6 +531,18 @@ export const ModelEditForm = <T extends StrapiModel>({
               {t('posts')}
             </ActionButton>
 
+            <ActionButton
+              isVisible={endpoint === 'notifications'}
+              onClick={e => {
+                sendNotificationButtonOnClick(e)
+              }}
```
7alip, line 538:
```suggestion
              onClick={sendNotificationButtonOnClick}
```

7alip (commented):
Thanks

7alip (approved):
I did some refactoring, which is unrelated to this PR.

I noticed that we use Next.js api routes for sending notifications. I'm not sure if it's the best place to handle this. 
I have some concerns:

1. Next.js serverless functions have some limitations about function size, request timeouts etc.
2. Obtaining all subscribers with API request might not be an efficient way. It's probably best to handle this in the Strapi server.
3. If we want to send notifications to role groups or specific users, we can just send the roles and user ids, then we can handle this logic in Strapi Controller more easily.

I created another task for notification forms, let's update this logic in there.

You can merge this.

PR #1443, opened 2024-06-17
join form email exist error.

myniqx:
image

can u extra check backend profile.create function in `apps/api/src/api/profile/controllers/profile.ts.`

### DIFF
--- a/apps/api/scripts/add-i18.sh
+++ b/apps/api/scripts/add-i18.sh
@@ -14,32 +14,32 @@ declare -a diller=("tr" "en" "nl")
 
 # Eğer default_value boş ise anahtarları ve değerleri görüntüleyelim
 if [ -z "$default_value" ]; then
-  for dil in "${diller[@]}" ; do
-    dosya="packages/ui/public/locales/${dil}/common.json"
-
-    # Dosya var mı kontrol edelim
-    if [[ -f "$dosya" ]]; then
-      echo ""
-      echo -e "${BOLD}> Processing file: ${GREEN}$dosya${RESET}"
-      echo "------------------"
-      # jq kullanarak key'i büyük küçük harfe duyarsız olarak içeren değerleri bulalım ve key=value şeklinde yazdıralım
-      jq -r --arg key "$key" 'to_entries | map(select(.key | test($key; "i"))) | .[] | "\(.key)=\(.value)"' "$dosya" | while IFS= read -r line; do
-        # echo -e ile doğrudan değiştirelim
-        echo -e "${line//$key/${RED}${BOLD}${key}${RESET}}"
-      done
-    fi
-  done
+    for dil in "${diller[@]}" ; do
+        dosya="packages/ui/public/locales/${dil}/common.json"
+        
+        # Dosya var mı kontrol edelim
+        if [[ -f "$dosya" ]]; then
+            echo ""
+            echo -e "${BOLD}> Processing file: ${GREEN}$dosya${RESET}"
+            echo "------------------"
+            # jq kullanarak key'i büyük küçük harfe duyarsız olarak içeren değerleri bulalım ve key=value şeklinde yazdıralım
+            jq -r --arg key "$key" 'to_entries | map(select(.key | test($key; "i"))) | .[] | "\(.key)=\(.value)"' "$dosya" | while IFS= read -r line; do
+                # echo -e ile doğrudan değiştirelim
+                echo -e "${line//$key/${RED}${BOLD}${key}${RESET}}"
+            done
+        fi
+    done
 else
-  # Her bir dil için işlemi tekrarlayalım
-  for dil in "${diller[@]}" ; do
-    dosya="packages/ui/public/locales/${dil}/common.json"
-
-    # Dosya var mı kontrol edelim, yoksa oluşturalım.
-    [[ -f "$dosya" ]] || touch "$dosya"
-
-    echo -e "${BOLD}> Processing file: ${GREEN}$dosya${RESET}"
-    # jq kullanarak anahtarı ve değeri ekleyelim
-    jq --arg key "$key" --arg value "$default_value" \
-      'to_entries | map(select(.key | test($key; "i") or .value | test($value; "i"))) | .[] | "\(.key)=\(.value)"' "$dosya" > "$dosya.tmp" && mv "$dosya.tmp" "$dosya"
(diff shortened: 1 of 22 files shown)


apps/api/src/api/profile/controllers/profile.ts
```diff
@@ -42,23 +46,41 @@ export default factories.createCoreController('api::profile.profile', () => {
...
-          where: { email: { $eq: email } },
-        })
+        .findOne({ where: { email: { $eq: email } } })
 
-      if (!profileResults || profileResults.length === 0) {
-        // no registered profile before
-        return super.create(ctx)
+      if (profile && !isRegistering) {
+        throw new ApplicationError('Profile already exists', {
+          code: 'strapi.error.create-profile.profile-already-exist',
```
7alip, line 74: `code` reminds me error status codes, such as `200`, `403` etc. We can use `key` or `i18nkey` instead maybe.

apps/api/scripts/add-i18.sh
```diff
@@ -14,32 +14,32 @@ declare -a diller=("tr" "en" "nl")
...
+        
+        # Dosya var mı kontrol edelim, yoksa oluşturalım.
+        [[ -f "$dosya" ]] || touch "$dosya"
+        
+        echo -e "${BOLD}> Processing file: ${GREEN}$dosya${RESET}"
+        
+        jq --arg key "$key" --arg value "$default_value" \
+        '. + { ($key): $value }' "$dosya" > "$dosya.tmp" && mv "$dosya.tmp" "$dosya"
+    done
 fi
```
7alip, line 45: Auto translated with Copilot

```bash
#!/bin/bash

key="$1"
default_value="${2:-}"

# ANSI codes for color and bold
RED='\033[0;31m'
GREEN='\033[0;32m'
BOLD='\033[1m'
RESET='\033[0m'

# Put the languages into an array
declare -a languages=("tr" "en" "nl")

# If default_value is empty, let's display the keys and values
if [ -z "$default_value" ]; then
    for lang in "${languages[@]}" ; do
        file="packages/ui/public/locales/${lang}/common.json"
        
        # Check if the file exists
        if [[ -f "$file" ]]; then
            echo ""
            echo -e "${BOLD}> Processing file: ${GREEN}$file${RESET}"
            echo "------------------"
            # Find the values containing the key case-insensitively using jq and print them in key=value format
            jq -r --arg key "$key" 'to_entries | map(select(.key | test($key; "i"))) | .[] | "\(.key)=\(.value)"' "$file" | while IFS= read -r line; do
                # Let's directly replace with echo -e
                echo -e "${line//$key/${RED}${BOLD}${key}${RESET}}"
            done
        fi
    done
else
    # Repeat the process for each language
    for lang in "${languages[@]}" ; do
        file="packages/ui/public/locales/${lang}/common.json"
        
        # Check if the file exists, if not, let's create it.
        [[ -f "$file" ]] || touch "$file"
        
        echo -e "${BOLD}> Processing file: ${GREEN}$file${RESET}"
        
        jq --arg key "$key" --arg value "$default_value" \
        '. + { ($key): $value }' "$file" > "$file.tmp" && mv "$file.tmp" "$file"
    done
fi
```

packages/lib/src/mutation/mutation.ts
```diff
@@ -115,6 +115,12 @@ export const mutation = async <
   } catch (error: any) {
     console.error('Mutation error', error)
 
+    // i dont know why but this way onError has first parameter
+    if (error.response?.data?.error?.details?.code) {
+      throw error.response?.data?.error?.details?.code
+    }
```
7alip, line 121: I wonder how frontend catches the error when we throw other error classes

```ts
const {
  ApplicationError,
  ForbiddenError,
  HttpError,
  NotFoundError,
  NotImplementedError,
  PaginationError,
  PayloadTooLargeError,
  PolicyError,
  RateLimitError,
  UnauthorizedError,
  ValidationError,
  YupValidationError,
} = utils.errors
```

apps/api/src/api/profile/controllers/profile.ts
```diff
@@ -42,23 +46,41 @@ export default factories.createCoreController('api::profile.profile', () => {
...
+          we need to check if email address has submitted a join form before
+          if so, we need to update attach the user to the profile
+          otherwise, we need to create a new profile than attach the user
+
+        in second case;
+          we need to create a new profile and if mail address has submitted a join form before
+          we need to return meaningful error        
+      */
+
+      const { email, user } = ctx.request.body.data
```
7alip, line 64: Do you think it's easier to maintain if we use a different controller for join form.

We use the `create` controller when users register from the Nextjs server. There we use the user's own jwt token. 

```ts
await Mutation.post('profiles', body, response.data?.jwt)
```

We can reach user context with this jwt token and we can associate the profile with this user.

I'm not sure if it's safe to take `user` data from the request body.

packages/lib/src/mutation/mutation.ts
myniqx: i tried ApplicationError and ForbiddenError. both are cought by this catch successfully. but in this catch, if we throw with new Error, mutate.onError catches it without error's parameter.

apps/api/src/api/profile/controllers/profile.ts
myniqx: i aggre that it ll be easy to maintain. but i do not think it will provide enough simplicity to save the overhead of creating a separate controller and making the necessary adjustments to all roles.
on this control, i dont think it will provide extra security. we get userId but we do not use it in the query, it is just a parameter that understands whether user signs up or not. but at the end it is your call, i can implement it.

myniqx (commented):
@7alip i pushed some of reviews.

packages/lib/src/mutation/mutation.ts
```diff
@@ -115,6 +115,12 @@ export const mutation = async <
   } catch (error: any) {
     console.error('Mutation error', error)
 
+    // i dont know why but this way onError has first parameter
+    if (error.response?.data?.error?.details?.i18nKey) {
+      throw error.response?.data?.error?.details?.i18nKey
+    }
+
+    // but this way, mutate.onError doesn't have first parameter
```
7alip, line 123: Alright, it's important point. We should keep this in mind when we use a global error catching

7alip (approved):
Thanks

PR #1447, opened 2024-06-21
added blogs tab to profile component

myniqx:
image

image


packages/ui/src/components/ProfileSettings/BlogsTab.tsx
```diff
@@ -0,0 +1,55 @@
...
+import { useGetAuthorBlogs } from '@fc/services'
+import { Blog } from '@fc/types'
+
+import { ModelEditForm } from '../../admin'
+import { BlogCard } from '../BlogCard'
+
+export const BlogsTab = () => {
+  const [selectedBlog, setSelectedBlog] = useState<Blog | null>(null)
+  const { t } = useTranslation()
+  const { data: blogs, refetch } = useGetAuthorBlogs()
```
7alip, line 15: I get 400 error probably because of the same error I had for arts.find controller

```ts
{
  "status": 400,
  "name": "ValidationError",
  "message": "Invalid parameter author",
  "details": {}
}
```

Author field is associated with `profile` endpoint and `author` role doesn't have a permission for profiles. We can override the find controller.

Which role did you use, admin?

7alip (commented):
Thanks

packages/ui/src/components/ProfileSettings/ProfilePanel.tsx
```diff
@@ -50,13 +51,17 @@ export const ProfilePanel: FC<ProfilePanelProps> = ({
...
+  const { user, profile, site } = useAuthContext()
   const orientation = useBreakpointValue<TabsProps['orientation']>({
     base: 'horizontal',
     lg: 'vertical',
   })
   const { t } = useTranslation()
 
+  const isBlogsVisible =
+    (site === 'dashboard' || site === 'foundation') &&
+    (user?.roles.includes('admin') || user?.roles.includes('Author'))
```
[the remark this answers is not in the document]
myniqx: image
that's why i added this colon :) corrent one is 'author'

packages/ui/src/components/ProfileSettings/BlogsTab.tsx
myniqx: but its public data. i mean all user without login they can see blogs from site. and this hook uses blogs endpoint which is the same one. also there is no fetching data from profile endpoint. it uses authContext hook to get profiles id which all user gets them when they login.

7alip: It's my mistake then. I see that there is a problem in the request url, probably because of strapiRequest function.

`http://localhost:1337/api/blogs?pagination[page]=1&pagination[pageSize]=25&filters[author][id][$eq]=38&locale=en&populate=%2A&sort[0]=publishedAt%3Adesc`

`&sort[0]=` part is not completed

7alip (approved):
Thanks

PR #1459, opened 2024-06-23
Solve 1,2 and 3. examples

SelimEmre:


packages/ui/src/examples/E2_FetchWithStrapiRequest.tsx
```diff
@@ -2,16 +2,35 @@
...
-    // Remember that fetcher is a wrapper around axios that adds the token and api url to the request
-  }, [])
+    const fetchData = async () => {
+      try {
+        const response = await strapiRequest<Blog>({
+          endpoint: 'blogs',
+          locale: 'tr',
+        });
+
+        if (response?.data?.length != 0) {
```
7alip, line 19: This condition will not behave as expected if response?.data?.length is undefined. Consider explicitly checking for both undefined and a non-zero length to make the intent clear and handle all cases correctly. For example:


```ts
if (response?.data?.length > 0) {
  setBlogs(response.data)
}
```

or

```ts
if (!response.data?.length) {
  return
}

setResponse(response.data)
```

7alip (commented):
Thanks

PR #1478, opened 2024-06-30
Fix datetime conversion to UTC before saving on hashtag page

ertugrul-aydin:
Changes
- Added conversion of datetime fields to UTC in the `onSaveModel` function.
- Ensured that the converted datetime is in ISO format before being sent to Strapi.

Testing
- The fix has been tested both locally and in production environments to ensure that the datetime fields are saved correctly.


ertugrul-aydin:
Changes
- Added a check in `mutation.ts` to identify Date instances in the request body.
- Converted identified Date instances to UTC and formatted them as ISO strings before sending the request.

7alip (approved):
Thanks

PR #1500, opened 2024-07-03
Mustafa

MustafaBudak41:


package-lock.json
7alip, line 1: You should always install packages with `yarn install` and never use `npm` as the package manager of this project is yarn. We shouldn't use both in a project.

7alip (commented):
Screenshot 2024-07-04 at 15 22 59

It seems you create your `mustafa` branch from `main` instead of `dev`. Otherwise we wouldn't see this commit history.

Always create a new branch from up to date dev branch.

7alip:
@MustafaBudak41 I'm closing this PR. Could you please create a new branch from `dev`?

PR #1502, opened 2024-07-04
added load more button.

myniqx:


packages/ui/src/components/AuditLogList/AuditLogList.tsx
```diff
@@ -36,10 +36,15 @@ export const AuditLogList: FC = () => {
             { uid: { $containsi: q } },
           ],
         },
-        pageSize: 100,
+        pageSize: 20,
```
7alip, line 39: Could be 50

7alip (approved):
Thanks

packages/ui/src/components/AuditLogList/AuditLogList.tsx
myniqx, line 39:
```suggestion
        pageSize: 50,
```

PR #1506, opened 2024-07-07
Email

7alip:


myniqx (approved):
may be we should add StrapiLocale column (which ll be updated when user change its locale) to Profile model, so we can send email to that profile accordingly.

PR #1527, opened 2024-07-18
Fix blog

7alip:


MustafaTRKYLMZ:
There is an error in blog detail page. Author is not coming to UI. And also in mobile version, when I click the edit button modal  seems so small. I saw this error in different pages. Activity detail pages something like that.

PR #1561, opened 2024-08-05
`Test` Create Account

FilizBUDAK:


packages/playwright/src/tests/creatingAccount.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+  await page.getByRole('link', { name: 'Kayıt ol' }).click() //02. The user clicks on the "Register" link on the page that opens.
+  await page.getByPlaceholder('İsim').click()
+  await page.getByPlaceholder('İsim').fill('Test') //03. The user enters the "name" text box.
+  await page.getByPlaceholder('Kullanıcı Adı').click()
+  await page.getByPlaceholder('Kullanıcı Adı').fill('testuser') //04. User enters a random name in the "Username" text box.
+  await page.getByPlaceholder('E-posta').click()
+  await page.getByPlaceholder('E-posta').fill('invalid-email') //05. User enters invalid email in the "Email" text box.
+  await page.getByPlaceholder('Parola').click()
+  await page.getByPlaceholder('Parola').fill('2Wsx.2Wsx.') //06. User enters their password in the "Password" text box.
+  await page.getByRole('button', { name: 'Hesap oluştur' }).click() //07. User clicks on "Create Account" button
```
7alip, line 70: This block is being used multiple times. That's why we use Page object model or helper functions. There could be a `Register.ts` page model like `Login.ts`. Consider having a `register` function there:

```ts
// Outside RegisterPage class
type RegisterArgs = {
  name: string,
  username: string,
  email: string,
  password: string,
}

async register({ name, username, email, password }: RegisterArgs) {
  // Code
}
```

> If it will be difficult to implement this, leave it simple as it is now.

packages/playwright/src/tests/creatingAccount.spec.ts
```diff
@@ -0,0 +1,91 @@
...
+  await expect(page.getByText('email must be a valid email')).toBeVisible() // Error message checking
+  await page.waitForTimeout(1000)
+})
+
+test('TC-0004 Negative flow (login with invalid information)', async ({
+  page,
+}) => {
+  await page.setViewportSize({ width: 1920, height: 1080 })
```
7alip, line 78: Is this `setViewportSize` required? I think we can add more devices in playwright config in the future if we want to test it for mobile or tablet.

7alip (commented):
Thanks. I left some comments.

I change the target branch from main to dev. It seems this branch is created from `main`. Could you create a new branch from `dev` called `test/create-account` and a new pull request?

PR #1594, opened 2024-08-19
`Test` dasbord verify

safaktopkaya:


packages/playwright/src/tests/dasbord-verifiy.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+  await page.getByRole('link', { name: 'Contact' }).first().click()
+  await page.getByRole('button', { name: 'EN', exact: true }).click()
+  await page.getByRole('link', { name: 'Arts' }).first().click()
+  await page.getByText('Other').click()
+  await page.getByRole('link', { name: 'Collections' }).first().click()
+  await page.getByRole('link', { name: 'Activities' }).click()
+  await page.getByRole('link', { name: 'About Us' }).first().click()
+  await page.getByRole('link', { name: 'Contact' }).first().click()
+
+  await page.locator('.css-1x01hl1').first().click()
```
7alip, line 44: You should avoid using random generated classes because it may be different in each deploy.

Your test can pass locally but it seems it fails in the pipeline: https://github.com/freedomcombination/monorepo/actions/runs/10460458698/job/28966665614?pr=1594

7alip (changes requested):
Thanks

packages/playwright/src/tests/dasbord-verifiy.spec.ts
```diff
@@ -0,0 +1,64 @@
...
+    page.getByRole('heading', { name: 'Sanat Durağı' }),
+  ).toBeVisible()
+  await expect(page.getByRole('link', { name: 'Eserlere Gözat' })).toBeVisible()
+  await page.getByRole('button', { name: 'TR' }).click()
+  await page.getByRole('heading', { name: 'Sanat Durağı' }).click()
+  await page.getByText('"Farklı renklerin buluşma').click()
+  await page.getByRole('link', { name: 'Eserlere Gözat' }).click()
+  await page.getByText('Kategoriler').click()
+  await page.getByRole('button', { name: 'NL' }).first().click()
+  await page.getByRole('link', { name: 'Kunsten' }).first().click()
```
7alip, line 31: @safaktopkaya Do you only check whether the menus are clickable and visible in the correct language? Or do you also want to see if the links work and take you to the correct page? If you want to test whether the link functionalities work, you should add assertions after each click.

For example:

```ts
await page.getByRole('link', { name: 'Kunsten' }).first().click()
await page.waitForURL(<expected url>, { timeout: TEST_TIMEOUT })
await expect(page).toHaveURL(<expected url>)
```

Alternatively, you can test if the page title displayed in the correct language

```ts
await page.getByRole('link', { name: 'Kunsten' }).first().click()
await page.waitForLoadState('networkidle')
const title = await page.title()
await expect(title).toBe(<expected title>)
```

safaktopkaya: My test case these are;o I tested them by following these steps

-username/language bar should be visible
-page title should be visible as Home-Kunsthalte
-buttom menu bar/ buttons should be active / view activities ,arts , collections clicable and 
 direct to other page

packages/playwright/src/tests/dasbord-verifiy.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+test('Dashbord-verify', async ({ page }) => {
+  await page.goto(getVercelUrl('kunsthalte'))
+
+  await page.getByRole('link', { name: 'Sign in' }).first().click()
+  await page.getByTestId('input-identifier').fill('admin')
+  await page.getByTestId('input-password').fill('Test?123')
+  await page.getByTestId('button-submit-login').click()
+
+  await expect(page.getByRole('button', { name: 'Admin' })).toBeVisible()
+})
```
7alip, line 14: Let's remove this test. We will add dashboard tests later in a separate task.

packages/playwright/src/tests/dasbord-verifiy.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+
+test('Language bar', async ({ page }) => {
```
7alip, line 16: Could you update your tests with the following format:

```ts
import { expect, test } from '@playwright/test'

test.describe('07. Verify Pages', () => {
  test('TC-01: should display homepage in the correct language', async ({
    page,
  }) => {
    // Test steps
  })

  test('TC-02: should display arts categories in the correct language', async ({
    page,
  }) => {
    // Test steps
  })
})
```

7alip (changes requested):
Could you rename your file name to `07-verify-pages.spec.ts`

packages/playwright/src/tests/dasbord-verifiy.spec.ts
safaktopkaya: Ican make update

PR #1596, opened 2024-08-20
Feat/empty-arts-result

bbluechip:


packages/ui/src/components/SearchForm/SearchForm.tsx
```diff
@@ -79,5 +81,11 @@ export const SearchForm: React.FC<SearchFormProps> = ({
         )}
       </InputRightElement>
     </InputGroup>
+    {isFetching && (
+      <p style={{ marginTop: '10px', color: 'red' }}>
```
myniqx, line 85: same here. but real problem is this is the common search component. i mean this component is not here to make search. its here to set a search parameter whereever its needed.

myniqx (changes requested):
```
           <SearchForm
                placeholder={t('search') as string}
                onSearch={value => changeParam({ searchTerm: value as string })}
                isFetching={artsQuery.isFetching}
              />
// as you can see, the component only changes props.
// onSearch={value => changeParam({ searchTerm: value as string })}
```

what you need is more about the search result. you are in the right track. go back to file 'ArtClubTemplate.tsx' the file that you found SearchForm component. and look a little below you ll see a '<MasonaryGrid' thats is where art results are listed. if you check those codes :
```
          {artsQuery.isLoading
                ? 
/* show something while loding */
                : 
/* show the result */
                  }
```
what you need to go investigate the code related 'show the result'...
if there is no data then show one of Chackra's component or if there is a custom one..
(another tip there is no custom component for this alert so you need to go to 
https://v2.chakra-ui.com/docs/components/alert#composition then you can see sample code below. copy it  paste it modify it :) )

packages/ui/src/components/SearchForm/SearchForm.tsx
```diff
@@ -25,7 +25,7 @@ export const SearchForm: React.FC<SearchFormProps> = ({
 }) => {
```
7alip, line 25: I don't think you should adjust SearchForm component for showing the empty result.

If you check `ArtClubTemplate`, there you could display empty result under the following condition:

```tsx
{!artsQuery.isLoading && artsQuery.data?.data?.length === 0 && <EmptyResultAlert />}
```

7alip (changes requested):
Thanks

apps/api/.strapi-updater 2.json
```diff
@@ -0,0 +1,5 @@
+{
+	"latest": "4.23.0",
+	"lastUpdateCheck": 1712850390421,
+	"lastNotification": 1709918503281
+}
```
7alip, line 5: This is not removed yet

apps/api/types/generated/components.d.ts
```diff
@@ -47,6 +31,22 @@ export interface CourseCurriculum extends Schema.Component {
   }
 }
 
+export interface FaqFaq extends Schema.Component {
```
7alip: I still don't know why Strapi changes this file everytime, you can revert this change from the git menu unless you modify the backend.

packages/ui/src/components/ArtClubTemplate/ArtClubTemplate.tsx
```diff
@@ -168,8 +170,11 @@ export const ArtClubTemplate: FC = () => {
                           recaptchaToken={recaptchaToken}
                         />
                       </AnimatedBox>
-                    )
-                  })}
+                    ); 
+                  }) :
+                  (
+                    <EmptyResultAlert/>
```
7alip, line 176: I typed in the comments `<EmptyResultAlert />` as an example. You do not need to create a new component for such a static component unless it is intended to be used in multiple places.

You can use this as an alert:

```tsx
<Alert
  status="warning"
  variant="subtle"
  flexDirection="column"
  alignItems="center"
  justifyContent="center"
  textAlign="center"
  height="200px"
  gap={4}
  rounded={"lg"}
>
  <AlertIcon boxSize="40px" mr={0} />
  <AlertDescription fontWeight={600} maxWidth="sm">
    {/* TODO: Translate */}
    No results found
  </AlertDescription>
</Alert>
```

packages/ui/src/components/ArtClubTemplate/ArtClubTemplate.tsx
```diff
@@ -154,7 +155,8 @@ export const ArtClubTemplate: FC = () => {
                       isMasonry
                     />
                   ))
-                : artsQuery.data?.data?.map((art, i) => {
+                : artsQuery.data?.data?.length as number > 0
```
7alip, line 158: If you render EmptyResultAlert in the grid component you won't have a full width alert container on desktop.

I would do something like:

```tsx
{
  !artsQuery.isLoading && artsQuery.data?.data?.length === 0 ? (
    <Alert />
  ) : (
    <MasonryGrid />
  );
}
```

packages/ui/src/components/EmptyResultAlert/EmptyResultAlert.tsx
```diff
@@ -0,0 +1,20 @@
...
+        return () => clearTimeout(timer); 
+      }, []);
+    
+      return showMessage ? (
+        <Alert status='info'>
+                      <AlertIcon />
+                        nothing was found
+                    </Alert>
+      ) : null;
+}
```
7alip, line 20: This component is not needed but I just give some feedbacks here:

1. Component does not need to be delayed, keep it simple.
2. To make the code more readable I usually prefer using early return both in components and functions:

```tsx
if (!showMessage) { return }

return <Alert />
```
3.  Your code is overall not formatted. The code should be formatted whenever you save. You might not have Prettier and ESLint plugins installed. Otherwise you should format it yourself. If you can't fix this auto formatting, just run `yarn format` before you commit.

7alip (changes requested):
Thanks

packages/ui/src/components/SearchForm/SearchForm.tsx
```diff
@@ -25,7 +25,7 @@ export const SearchForm: React.FC<SearchFormProps> = ({
...
   const [searchTerm, setSearchTerm] = useState('')
   const [debouncedSearchTerm, setDebouncedSearchTerm] = useState('')
-
+  // State that controls if any results after search
```
7alip, line 28: The changes in this `SearchForm` should also be reverted.

myniqx:
there are alot of common.json file you need to modify to get it work.
there are commands for them to make it work easy.

for example if you use 
`yarn add-word "key-for-your-sentence" "your sentence"`  example: `yarn add-word no-result-found "No result found"`
this ll add public>common files...
if you use it like `yarn add-word result` without last parameter, it ll list all keys included 'result' so you can use it instead of re-adding same sentence.
then `yarn copy-locales` to copy public locale files to all other places. 

also there is translate page on dashboard that appears only if you are in development mode which can easly translate your added keys to other language. if you use that page to translate your keys you can skip `yarn copy-locales` because when you save it, it saves to all places. btw you need to restart application to see effects or you ll see only keys (like 'no-result-found')

7alip (approved):
Thanks, you can merge it

PR #1600, opened 2024-08-25
Feat/password-validation-update

bbluechip:


packages/ui/src/components/ProfileSettings/ChangePasswordForm.tsx
```diff
@@ -11,7 +11,7 @@ import { API_URL } from '@fc/config'
 import { useAuthContext } from '@fc/context'
 
 import { FormItem } from '../FormItem'
-
+// mark
 const schema = Yup.object().shape({
   currentPassword: Yup.string().required(),
   password: Yup.string().required(),
```
7alip, line 17: `password` field has some rules in SignupForm but here it's only `required` without any extra rule.

```ts
password: yup
      .string()
      .min(8, t('login.password.warning', { count: 8 }) as string)
      .required()
      .matches(
        RegExp('(.*[a-z].*)'),
        t('login.password.matches.lowercase') as string,
      )
      .matches(
        RegExp('(.*[A-Z].*)'),
        t('login.password.matches.uppercase') as string,
      )
      .matches(
        RegExp('(.*\\d.*)'),
        t('login.password.matches.number') as string,
      ),
```

packages/ui/src/components/SignupForm/SignupForm.tsx
```diff
@@ -26,7 +26,7 @@ import { SignupFormFieldValues, SignupFormProps } from './types'
 import { ButtonLink } from '../ButtonLink'
 import { FormItem } from '../FormItem'
 import { SocialLoginButtons } from '../SocialLoginButtons'
-
+// mark
```
7alip, line 29: Remove unnecessary comment here. Only ChangePasswordForm would be adjusted.

7alip (changes requested):
Thanks

7alip (approved):
Thanks

PR #1617, opened 2024-08-31
added total price to asset

### LINKED ISSUE #1428: FEAT: Develop a price calculation for assets.

myniqx:
image

resolves #1428

### DIFF
--- a/packages/ui/src/components/DataTables/DataTable.tsx
+++ b/packages/ui/src/components/DataTables/DataTable.tsx
@@ -1,6 +1,6 @@
-import { useMemo } from 'react'
 
 import {
+  Badge,
   Box,
   HStack,
   Image,
@@ -8,12 +8,11 @@ import {
   Spacer,
   Stack,
   Text,
-  VStack
+  VStack,
 } from '@chakra-ui/react'
 import { useTranslation } from 'next-i18next'
 
-import { Asset, StrapiModel } from '@fc/types'
-import { formatPrice } from '@fc/utils'
+import { StrapiModel } from '@fc/types'
 
 import { DataTableProps } from './types'
 import { Pagination, WTable } from '../../components'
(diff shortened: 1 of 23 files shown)


packages/ui/src/components/DataTables/DataTable.tsx
```diff
@@ -25,10 +28,21 @@ export const DataTable = <T extends StrapiModel>({
   pageSize,
   setPageSize,
   allowExportPDF = false,
+  isAsset = false,
```
7alip, line 31: Could we make this more generic with an option prop? We might use this in other tables in the future.

```ts
totalOptions?: {
  label: string
  key: keyof T
  transform?: (value: number) => string | number
}
```

Then 

```ts
  const totalValue = useMemo(() => {
    if (!totalOptions) return null

    const items = tableProps.data as T[]

    if (!items || !items.length) return null

    return items.reduce((prev, curr) => prev + curr[totalOptions.key], 0)
  }, [totalOptions, tableProps.data])
```

Then

```tsx
<Text noOfLines={1} fontWeight={'bold'}>
  {label}{transform?.(totalValue) || totalValue}
</Text>
```

7alip (commented):
Thanks

packages/ui/src/components/DataTables/types.ts
```diff
@@ -13,4 +20,5 @@ export type DataTableProps<T extends StrapiModel> = {
   children?: ReactNode
   setPageSize: (pageSize: number) => void
   allowExportPDF?: boolean
+  badges?: DataTableBadgeProps<T> | DataTableBadgeProps<T>[]
```
7alip, line 23: Could always be an array `DataTableBadgeProps<T>[]`

packages/ui/src/components/DataTables/DataTable.tsx
```diff
@@ -71,6 +76,21 @@ export const DataTable = <T extends StrapiModel>({
...
+                  variant="outline"
+                  size={'2xl'}
+                  fontSize={'md'}
+                  p={1}
+                  colorScheme={'primary'}
+                  {...badge.badgeProp}
+                >
+                  {badge.badgeText(tableProps.data)}
+                </Badge>
+              ))}
```
7alip, line 93: Could be simpler if badges is always an array.

```ts
{badges?.map((badge, index) => {})
```

7alip (approved):
Thanks

PR #1619, opened 2024-08-31
Remove transformWithModel

7alip:


myniqx:
Doesnt yarn tsc go crazy and gives a thousand error on your pc after those lines?

7alip:
> Doesnt yarn tsc go crazy and gives a thousand error on your pc after those lines? Because i did the same and got those errors. @7alip

No, if tsc fails, Vercel must also be failed.

myniqx (approved):
> No, if tsc fails, Vercel must also be failed. Maybe your second parameter was not optional?

i didnt make second parameter optional on purpuse, i though that it ll be there allways. 
but this is the way. respect..

PR #1644, opened 2024-09-07
added toast to security tab functions

myniqx:


apps/dashboard/public/locales/tr/common.json
```diff
@@ -449,4 +452,4 @@
   "weeksWithCount_other": "{{count}} hafta",
   "welcome": "Hoşgeldin",
   "you-must-logged-in": "Eser yükleyebilmek için giriş yapmalısınız!"
-}
+}
```
7alip, line 455: An empty line at the end of the file is default behavior of prettier. Could you run `yarn format` or fix this problem inside your translations cli?

7alip (approved):
Thanks

PR #1669, opened 2024-09-12
Reset Pagination After Applying Filters

bbluechip:


bbluechip:
in my last commit, code works on some cases... Page numbers reset if selected category contains more than one page and if you are on that page before you choose that category.

packages/ui/src/components/CategoryFilter/CategoryFilter.tsx
```diff
@@ -49,10 +52,20 @@ export const CategoryFilter: FC<CategoryFilterProps> = ({
   useEffect(() => {
     setIsLoading(true)
   }, [value, setIsLoading])
+  
+  const resetPagination = () => {
```
7alip, line 56: It's not only about changing categories. `selectCategories` prop calls changeParam function.
This should be handled within the useChangeParams hook to ensure consistency across all query params updates.

I mentioned here: https://github.com/freedomcombination/monorepo/pull/1670#discussion_r1757455570

7alip (changes requested):
Thanks

packages/ui/src/hooks/useChangeParam.tsx
```diff
@@ -16,6 +16,10 @@ export const useChangeParams = () => {
     const newQuery = cleanQuery(query, args)
     cleanArgs(args)
 
+    if (args['categories']) {
+      args['page'] = 1
+    }
+
```
7alip, line 22: Not only `categories` filter but every possible filter other than `page` should be considered. For example, should the `page` filter should also be removed when people search `(?q=search)`. There can be other filters too.

7alip (changes requested):
It's not based on the latest dev. There will be conflicts when you try to merge `dev` into your branch.

I would create a new branch from dev.

PR #1676, opened 2024-09-15
i18help

myniqx:
`yarn i18nhelp`

here instructions :
https://github.com/myniqx/i18n-help

and also `yarn i18nhelp help` gives guideness.
also if user enter deepL key, it can translate phrases to all locales automatically while adding.


7alip (approved):
Thanks for your efforts

PR #1690, opened 2024-09-17
Reset Pagination After Applying Filters

bbluechip:


packages/ui/src/hooks/useChangeParam.tsx
```diff
@@ -60,13 +60,15 @@ export const useChangeParams = () => {
       const sanitizedArgs = sanitizeArgs(args)
 
       // TODO: Remove existing page query if other params are changed
-
+      if (args['categories']) {
```
7alip, line 63: Again, it's not based on the latest dev. `useChangeParams` had been updated. Compare it with dev. It seems you created your new branch without updating the dev.

Screenshot 2024-09-17 at 20 36 23

7alip (changes requested):

PR #1830, opened 2024-10-25
some changes on form-item

myniqx:
this is the approach i use to use useContext with FormItem. If it is okey, i will update the places used in the project as in LoginForm.


7alip:
@myniqx Let's wait for the chakra branch to be merged.

PR #1850, opened 2024-10-28
fixed false already registered notification.

myniqx:
it seems somehow both comes undefined and find return a profile.


7alip (approved):
Thanks, you can merge if it's already formatted

PR #1855, opened 2024-10-29
yet another patch for dates.

myniqx:
image

it didnt show but im editing and saving continuesly.


packages/ui/hooks/useDefaultValues.ts
```diff
@@ -51,48 +51,69 @@ export const useDefaultValues = <T extends StrapiModel>(
...
 
-    const getDate = (date?: string | null, isDateTime?: true | 'amsterdam') => {
+    const convertDate = (
+      fieldName: string | number | symbol,
+      toDateOnly: boolean,
+      date?: string | null,
+    ) => {
       if (!date) return ''
 
       const dateTime = new Date(date)
```
7alip, line 62:
```suggestion
      const dateTime = parseISO(date)
```

Could you try parseISO if it fixes timeout problems without needing timezone difference calculation?

Otherwise you can merge it.

7alip (commented):
Thanks

PR #1862, opened 2024-10-30
Fix sending emails from dev/staging

7alip:


apps/api/src/api/donate/controllers/custom.ts
```diff
@@ -8,6 +8,11 @@ export default {
       {},
     )
 
+    if (process.env.NEXT_PUBLIC_ENVIRONMENT !== 'production') {
+      return { message: 'Email sent (Dev mode)' }
+    }
+
+    // TODO: Use sendReactEmail
```
7alip, line 15: > There is not much to review here. But there is a if condition in sendreactmail which checks environment. We need to change it. It only uses devmail in development. And all other place use sendreactmail without checking environment. But still sendreactmail requires a react mail template.

@myniqx We receive a lot of emails because of automation tests. I put these comments not to forget to make new templates for each scenario.

7alip: I know we have 2 years old TODO comments :) But it's better than adding nothing. There are already general email templates tasks to be prepared.

PR #1871, opened 2024-11-02
parseISO

myniqx:
when i test it, it seems to work with dates(but i doubt its because of paseISO, i still believe its because of yup.date()) but not with date-times. hours are switching when we save. i publish this anyway. maybe you want to give a try as well.


7alip:
@myniqx yup.date is annoying. I was considering to switch to a separate time field for datetime values in the future. For now it's probably merge https://github.com/freedomcombination/monorepo/pull/1855

https://react-spectrum.adobe.com/react-aria/TimeField.html

PR #1873, opened 2024-11-02
Fix empty myApplication

7alip:
Course application details modal is broken probably because of empty `myApplication` field. Can you check what can cause that?


myniqx:
depends on the location. for example if user looking its profile's course's page and if there are some courses so it must be defined. basicly it does :  allApplications.filter( if applicant is my profile ) then list all courses to my profile. i just checked it on my local ( btw i still didnt update with mock database ) so it runs without error. maybe the problem is in that mock data.

7alip:
@myniqx I mean, you may have updated the profile-related permissions required for the course application, but they may not have been updated in production yet. If it's permissions related of course.

PR #1878, opened 2024-11-03
Refactor date time

7alip:


packages/services/common/mutation.ts
```diff
@@ -137,9 +137,9 @@ export const mutation = async <
 
     Object.entries(body).forEach(([key, value]) => {
       if (value instanceof Date) {
-        const timeDif = getMinuteDifferenceAmsterdamBetweenLocal(value)
-        const amsterdamDateTime = addMinutes(value, -timeDif)
-        body[key as keyof D] = amsterdamDateTime.toISOString() as D[keyof D]
+        const localTimeOffset = getLocalTimeOffset(value)
```
myniqx, line 140: if we move this code to ModelEditForm.tsx onSaveModel, isnt it better? i didnt move before, bcs i dont know why this is added in this place (maybe there is a reason?)

myniqx (approved):

packages/services/common/mutation.ts
7alip: We use mutation also in create form and it could also be used in some custom mutation functions in the future. I just wanted to modify datetime from a single source.

PR #1891, opened 2024-11-04
added some missing locales. use chakra's For and Show as much as poss…

myniqx:
…ible to improve readability. fixed blog and arts tab (in case of no content). improved profile-tabs. and clear some lints.


7alip (approved):
Thanks

PR #1902, opened 2024-11-06
added prison fetch functionality, removed badge from data-table and a…

myniqx:
…dded panel which is more flexible. updated data-table's select to chakra v3's.


7alip:
@myniqx There are conflicts need to be resolved

myniqx:
@7alip okey its completed.

PR #1905, opened 2024-11-07
Update Fetch with axios

adamagar:
Please review the changes for Fetch with axios


packages/ui/examples/E1_FetchWithAxios.tsx
```diff
@@ -2,24 +2,65 @@
...
+
 // You can use local API_URL instead of the above url
 // yarn --cwd apps/api dev to run the api locally (http://localhost:1337)
 const BLOG_URL = `${API_URL}/api/blogs`
+const TOKEN = process.env.NEXT_PUBLIC_API_TOKEN || ''
 
-export const FetcWithAxios = () => {
-  const [blogs, setBlogs] = useState([])
+export const FetchWithAxios = () => {
+  const [blogs, setBlogs] = useState<Blog[]>([])
```
7alip, line 24: Nice that you indicate state type explicitly. I should have mentioned that you could import backend model types from our local types package, `@fc/types`:

`import type { Blog } from '@fc/types'`

packages/ui/examples/E1_FetchWithAxios.tsx
```diff
@@ -2,24 +2,65 @@
...
+        headers: {
+          Authorization: `Bearer ${TOKEN}`,
+        },
+      })
+
+      const blogsData: Blog[] = Array.isArray(response.data.data) ? response.data.data : []
+
+      if (blogsData.length === 0 && locale !== 'tr') {
+        const fallbackResponse = await axios.get(BLOG_URL, {
+          params: { locale: 'tr' },
```
7alip, line 43: > // NOTE: Not every locale may have a blog, so you may need to change the locale to tr

In normal page, you can just change the url `https://localhost:<port>/<locale>`. But for stories, you could change the language from the Storybook toolbar. It should update the value of `useRouter().locale`.

This fallback approach is nice though.

7alip (approved):
Thanks

Issue #1901, opened 2024-11-06
FEAT: Publish Playwright Reports to Github Pages


7alip:

bluewave07:
https://github.com/freedomcombination/monorepo/pull/1910

PR #1910, opened 2024-11-09
Akadir/playwright report

bluewave07:


.gitignore
```diff
@@ -49,7 +49,6 @@ storybook-static
 
 #playwright
 test-results
-playwright-report
```
7alip, line 52: Playwright generates report files after each test run. We should definitely ignore report files.

https://github.com/freedomcombination/monorepo/actions/workflows/playwright.yml

Screenshot 2024-11-09 at 22 26 38

If you create this PR because of https://github.com/freedomcombination/monorepo/issues/1901, the goal is to deploy generated report files to Github pages for quick viewing.

7alip (changes requested):
Thanks

PR #1916, opened 2024-11-10
in endpoint.tsx useChangeParams fully used. and some other additions.

myniqx:
if the changes about assets and courses endpoint is met requirements i can delete index to delete files.


apps/dashboard/pages/[endpoint].tsx
```diff
@@ -47,15 +47,15 @@ type ModelPageProps = InferGetServerSidePropsType<typeof getServerSideProps>
 
 const ModelPage: FC<ModelPageProps> = ({ endpoint }) => {
   const { t } = useTranslation()
-  const { locale, query, push } = useRouter()
+  const { locale, push } = useRouter() // Removed 'push' as it's not used
```
Comment (7alip) [inline apps/dashboard/pages/[endpoint].tsx:50]: What about this comment: // Removed 'push' as it's not used

apps/dashboard/pages/[endpoint].tsx
```diff
@@ -167,54 +167,9 @@ const ModelPage: FC<ModelPageProps> = ({ endpoint }) => {
...
-      (key === 'profileStatus' && value === 'all') ||
-      (key === 'published' && value === 'all') ||
-      (key === 'pageSize' && value === 20)
-    ) {
-      const _query = { ...query }
-      delete _query[key]
-      push({ query: _query }, undefined, { shallow: true })
-
-      return
-    }
```
Comment (7alip) [inline apps/dashboard/pages/[endpoint].tsx:195]: How do you handle these default values? Query params can be removed if the default value is set.

> NOTE: I was not happy with adding and extra `all` to the radio menu. It would be nice to have radio menu items that support de-selecting. We wouldn't have to add additional `all` item. Alternatively, we could have used `Clear filters` button but it's not as flexible as **deselecting** the item. Users can deselect a particular filter instead of clearing all the filters. We can improve it later.

7alip (approved):

7alip:
>if the changes about assets and courses endpoint is met requirements i can delete index to delete files.

@myniqx you can delete them

PR #1929, opened 2024-11-11
Mlalek/onboard 02

MLalek:
Onboarding 02 Test-write


MLalek:
Pls. review the task

playwright/onboarding/02-write-test.spec.ts
```diff
@@ -13,9 +13,16 @@ import { test } from '@playwright/test'
...
+test('Example', { tag: ['@onboarding'] }, async ({ page }) => {
   // TODO: Go to the URL `https://playwright.dev` and verify the page title contains 'Playwright'
   // (For example title is: "Fast and reliable end-to-end testing for modern web apps | Playwright") and contains 'Playwright'
+  await page.goto('https://playwright.dev/')
+  await page.pause
+  await page
+    .locator(
+      '//*[@id="__docusaurus_skipToContent_fallback"]/header/div/h1/span',
+    )
+    .isVisible()
```
7alip, line 25: What is `page.pause` for here?

There is no assertion here, without `expect` this test will always pass. Visibility test could be something like this:

```ts
const title = page.locator('<page-selector-here>')

// I don't recommend isVisible
const isTitleVisible = await title.isVisible()
await expect(isTitleVisible).toBe(true)

// I recommend toBeVisible over isVisible
await expect(title).toBeVisible()
```

But, even if you would check the visibility of the title, it would not cover the criteria.

---

I don't see any assertion for checking if the title contains `Playwright`.

- [ ] Assign the locator to a variable `const title = ...`
- [ ] Check if the locator contains (`Playwright`)

playwright/onboarding/02-write-test.spec.ts
```diff
@@ -30,6 +37,18 @@ test.describe('https://github.com', { tag: ['@onboarding'] }, () => {
...
+  test(
+    'should display page title',
+    { tag: ['@onboarding'] },
+    async ({ page }) => {
+      await page.goto('https://github.com/')
+      await page
+        .locator(
+          'xpath=/html/body/div[1]/div[1]/header/div/div[2]/div/nav/ul/li[1]/button',
+        )
+        .isVisible()
```
7alip, line 49: Again, this code does nothing here. This whole block means only `true` value. And it's not being used for anything.
Wouldn't you check if the title has the correct text (Whatever you see on the website)

```ts
const title = page.locator("")
expect (title).toBe("<Exact title content>")
  
// Alternatively
const titleContent = await title.textContent()
expect(titleContent).toBe("<Exact title content>")
```

playwright/onboarding/02-write-test.spec.ts
```diff
@@ -38,4 +57,18 @@ test.describe('https://github.com', { tag: ['@onboarding'] }, () => {
...
+    { tag: ['@onboarding'] },
+    async ({ page }) => {
+      await page.goto('https://github.com/')
+      await page.locator('//button[contains(@aria-label,"Search")]').click()
+      await page
+        .locator(
+          '//*[@id="query-builder-test-result-pricing"]/a/span[2]/span/span',
+        )
+        .click()
+      await page.locator('#nav-bar :text("Pricing")').isVisible
```
7alip, line 71: Same comment here

7alip (changes requested):
Thanks. Could you update your test in the same branch?

playwright/onboarding/02-write-test.spec.ts
MLalek: I made the necessary changes.

playwright/onboarding/02-write-test.spec.ts
MLalek: I made the necessary changes.

PR #1934, opened 2024-11-12
Fix uploading avatar

7alip:


packages/ui/components/ProfileSettings/AvatarForm.tsx
```diff
@@ -0,0 +1,145 @@
...
+  const { t } = useTranslation()
+  const { open, onOpen, onClose, onOpenChange } = useDialog()
+  const [file, setFile] = useState<File | Blob | null>(null)
+
+  const onCancel = () => {
+    setFile(null)
+    onClose()
+  }
+
+  const { mutate, isPending } = useUpdateModelMutation('profiles')
```
myniqx, line 38:
```ts
// in DetailsTab.tsx line 156
 type PartOfProfile = Pick<
    Profile,
    'name' | 'bio' | 'occupation' | 'phone' | 'city'
  >
  const { data, setData, hasChanged, isPending, onSave } =
    useProfileMutate<PartOfProfile>({
      name: profile?.name ?? null,
      bio: profile?.bio ?? null,
      occupation: profile?.occupation ?? null,
      phone: profile?.phone ?? null,
      city: profile?.city ?? null,
    } satisfies PartOfProfile)
```

you can consider to use useProfileMutate instead of useUpdateModelMutation. in a few tab, we use those mutate functions over and over so i wrote that useProfileMutate. it does all those repetative functions. all you need to do pass those props to components.

PR #1942, opened 2024-11-14
adding middleware to secure profile endpoint.

myniqx:
```ts
filterEndpoints(ctx, {
      "profile": {
        filtersByRoles: [
                 { roles: ['admin', /* you can add other roles with same filter */], only: '*' },
                 /* or you can add other roles with different filter */
        ], 
        filterToRequestOwner: {  // if (isRequestOwner is true) this filter applies
                // there are 2 options exclude or only..
                exclude: ['comment', 'observations']   // this ll removes if those field exist
        },
        isRequestOwner: (profile: { id: unknown }) => profile.id === profileId, // i called it requestowner but it can be anything
        filterToRest: { // if false this one applies
             only: ['name', 'city'] // this ll removes everything except 'name' and 'city', beside array it can be '*'...
         } 
      },
      // add as much as endpoint here as you need
    })
```


myniqx:
@7alip it was middlewares profileId query. when userId is undefined (such queries as token is not given) strapi gives that error.

myniqx:
@7alip it now auto-detects all fields.

7alip (approved):

PR #1943, opened 2024-11-15
Refactor secure filter

7alip:
I updated type and function names and provide generic types. There still some points we should consider:

- [ ] Update custom logic in `arts.find` controller
- [ ] Update custom logic in `blogs.find` controller
- [ ] Check all the controllers if there are custom implementations based on profile relation


apps/api/src/utils/filterResults.ts
```diff
@@ -1,82 +1,96 @@
+import { StrapiEndpoint, StrapiModel } from '@fc/types'
 import { checkRoles } from './checkRoles'
 
-type FilterOnly = string[] | '*'
-type FilterExclude = string[]
+type IncludedColumns<T extends StrapiModel> = Array<keyof T> | '*'
```
myniqx, line 4: It seems like the 'include' word does not fully meet its meaning. I recommend at least the name `SelectedColumns`.

apps/api/src/utils/filterResults.ts
```diff
@@ -1,82 +1,96 @@
...
+  data: T,
+  columns: IncludedColumns<T>,
+) => {
+  if (columns === '*') return data
 
-  return only.reduce((acc, col) => {
-    if (data[col]) acc[col] = data[col]
+  return columns.reduce((acc, col) => {
+    if (data[col]) {
+      return {
```
myniqx, line 30: i do a little asking to gbt to confirm. since reduce's acc is meant to be mutable (so we can modify safely), using spread operator can reduce performans with big objects.

and i cant stop diging :) (im sharing this to general information :)

```js
// { ... spread }
const obj = { a: 1, b: 2 };
const newObj = Object.assign({}, obj, { c: 3 }); // This is how TypeScript converts it to JavaScript

// And Object.assign looks like this:
if (typeof Object.assign !== 'function') {
  Object.assign = function (target, ...sources) {
    if (target == null) { // Check for null or undefined
      throw new TypeError('Cannot convert undefined or null to object');
    }

    // Force the target to become an object (object coercion)
    let to = Object(target);

    // Iterate over each source object
    for (let source of sources) {
      if (source != null) { // Skip null or undefined sources
        for (let key in source) {
          if (Object.prototype.hasOwnProperty.call(source, key)) {
            to[key] = source[key]; // Copy the property to the target
          }
        }

        // Copy symbol properties
        let symbols = Object.getOwnPropertySymbols(source);
        for (let symbol of symbols) {
          if (Object.prototype.propertyIsEnumerable.call(source, symbol)) {
            to[symbol] = source[symbol];
          }
        }
      }
    }

    return to;
  };
}

// assign[key] = value
const obj = { a: 1 };
assignProperty(obj, 'b', 2);
console.log(obj); // { a: 1, b: 2 }

function assignProperty(obj, key, value) {
  if (obj == null) {
    throw new TypeError('Cannot set property on null or undefined');
  }

  const propertyKey = String(key); // Convert the key to a string
  obj[propertyKey] = value;        // Assign the value to the key
}

```

7alip: Thanks :)

myniqx:
I merge it then continue from original branch. I have an idea to dynamically add profile fields to function.

PR #1949, opened 2024-11-15
`Test` Notification Modal

MustafaBudak41:
I couldn't use locater either because there were 3 tests named "language-switcher-tr".


7alip:
> I couldn't use locater either because there were 3 tests named "language-switcher-tr".

I updated the locators.

MustafaBudak41:
Testids added

PR #1957, opened 2024-11-17
TC-04: should be able to view the notifications

hilalnigarbasan:
TC-04: should be able to view the notifications.
- When I clicked on the notifications, no window opened, so I couldn't continue this test.

TC-05: should be able to view the subscription button.
- And I saw that the subscription button was deleted for this testcase that I wrote.

Screenshot 2024-11-17 at 01 17 15


7alip (commented):
Thanks

7alip:
Sorry I didn't notice that you updated the code. Let's merge this

PR #1967, opened 2024-11-17
filter added

MustafaTRKYLMZ:
- [ ] filter added
- [ ] fix [id] page


7alip:
@MustafaTRKYLMZ I handled it in #1965

PR #2023, opened 2024-11-21
Burak multi select

myniqx:
relation filter and boolean filter, both of them are multiselect and using changeParams.
endpointtable props are now generic. typescript gives you error etc if you didnt set correct filter.
tidied code a little. removed comment from endpointtable.


7alip (approved):
I don't have time to review it but functionality works perfect, excellent work!

PR #2051, opened 2024-11-23
Onboarding/hakanhvtc/02 write test

hakanhavutcu:


playwright/onboarding/02-write-test.spec.ts
```diff
@@ -30,6 +34,14 @@ test.describe('https://github.com', { tag: ['@onboarding'] }, () => {
   // - [ ] Update the test title to "should display page title"
   // - [ ] Goto the URL https://github.com
   // - [ ] Verify the first title element (h1) has the correct text
+  // Test code here
+  test('should display page title ', async ({ page }) => {
+    await page.goto('https://github.com')
+    const h1Locator = page.locator('h1').nth(0)
```
7alip, line 40: If the page structure changes, using a generic tag selectors with order can case tests to fail. I would consider using more specific selectors like an id, class or other attributes to target the element more reliably, if available.

7alip (approved):
Very good, thanks

7alip:
@bluewave07 `02 Write test` can be moved to Done

